Issue #13, opened 2017-08-15
Crash in [GADMAdapterMoPub loadNativeAdImages]


popic28:
We have a 300 crashes per day coming from the mopub adapter framework, and after checking the stacktrace the are coming from `GADMAdapterMoPub`'s `loadNativeAdImages` method. We've found that the crash is caused by inserting a nil object into an array. I'll open a PR with a fix (just a nil check before adding the URL to the images array)

ericleich:
Commentary is in #14 . Basically, we're working with MoPub on getting to the bottom of this; we think the issue may be malformed urls.

yood:
+1.

I'm getting a stream of crashes everyday at `-[GADMAdapterMoPub loadNativeAdImages]` with the error:

>    Fatal Exception: NSInvalidArgumentException
>    *** -[__NSArrayM insertObject:atIndex:]: object cannot be nil

I'm using these pods:
- Firebase/AdMob (4.8.0):
    - Firebase/Core
    - Google-Mobile-Ads-SDK (= 7.27.0)
- GoogleMobileAdsMediationMoPub (4.19.0.0):
    - Google-Mobile-Ads-SDK (>= 7.14.0)
    - mopub-ios-sdk/Core (= 4.19.0)

popic28:
It's 6 months later and still the issue isn't resolved, and the solution 

> Commentary is in #14 . Basically, we're working with MoPub on getting to the bottom of this; we think the issue may be malformed urls.

definitely did not pan out. 
Regardless of the malformed URL's or anything, the issue itself is about using an API that you don't want to use as it's vendor intends, namely the NSURL and NSArray API's. the breakdown goes like this : 
1. [NSURL URLWithString] says that it can return nil 
2. [NSMutableArray addObject:] will crash for object = nil

So when you want to write :
NSURL *URL = [NSURL URLWithString:_nativeAd.properties[key]]; 
[imageURLs addObject:URL];
you have to check that URLWithString did not return nil. 
If you want to 'fail early' you can add an assertion there, and handle the case gracefully in production instead of just crashing because 'the URL was malformed' and especially if what you're shipping is a third party ads framework that tons of live apps rely on. 

I hope the issue gets reopened or fixed soon.

bitomule:
Any update on this? We're still having the same issue.

maninomula:
I think the changes are live. Please take a look and get back to us.

bitomule:
[...]

PR #14, opened 2017-08-15
fix nil object in images array crash

popic28:
Hello all,

This PR adds a nil check after creating the image URL's.
I've reported the crash in issue #13 

Thanks, and let me know if there's anything else to do from my side.

### DIFF
--- a/adapters/MoPub/MoPubAdapter/GADMAdapterMoPub.m
+++ b/adapters/MoPub/MoPubAdapter/GADMAdapterMoPub.m
@@ -229,7 +229,8 @@ static NSString *const kAdapterErrorDomain = @"com.mopub.mobileads.MoPubAdapter"
         && [[_nativeAd.properties objectForKey:key] isKindOfClass:[NSString class]]) {
       if ([_nativeAd.properties objectForKey:key]) {
         NSURL *URL = [NSURL URLWithString:_nativeAd.properties[key]];
-        [imageURLs addObject:URL];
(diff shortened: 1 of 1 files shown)


popic28:
signed the CLA

ericleich:
Thanks for the PR. We'll take a closer look tomorrow.

Initial thoughts:
1) Are you able to reproduce locally? If so, can you share repro steps? I'm curious what the string is that would cause the URL to be nil. Nil check addresses the crash but it may be a larger problem that certain image assets aren't returned.
2) The PR itself seems to be missing a closing )

ericleich:
Turns out we are already working with MoPub on getting to the bottom of this issue. Current diagnosis is that the URLs come in malformed, but we can't reproduce.

We aren't sure that nilling out is the right solution, so we don't plan to accept this PR at this moment. If you think this change makes sense for you right now in your own app, feel free to make a build for yourself with these changes. The xcodeproj for the MoPub adapter should be set up to let you run the Fat adapter target after dropping in the Google Mobile Ads and MoPub libraries in [this directory](https://github.com/googleads/googleads-mobile-ios-mediation/tree/master/adapters/MoPub/Library).

ericleich:
Having said all that, if the issue is indeed malformed urls, the nil check wouldn't even prevent crashes...

sbagadi:
Fixed by https://github.com/googleads/googleads-mobile-ios-mediation/pull/19

PR #20, opened 2017-09-19
AdColony 3.2.1 support

shawnkc:
Also added new "test mode" and automatic attribution of mediation network within app options.


adapters/AdColony/Public/Headers/GADMAdapterAdColonyExtras.h
```diff
@@ -17,4 +17,7 @@
 /// Enables reward dialogs to be shown after an advertisement.
 @property BOOL showPostPopup;
 
+/// Enables test ads for your application without changing dashboard settings.
+@property BOOL testMode;
```
Rajaprahlad: This extra is not being used by the adapter. I think you might missing one commit.

shawnkc: You're right, thanks! I had a bad merge somewhere.

PR #30, opened 2017-11-13
Added nend Adapter files

fan-t-kinami:


fan-t-kinami:
Added banner and interstitial adapter.

PR #37, opened 2017-12-21
 Updated myTarget adapter

aserrr:
- Verified compatibility with myTarget SDK v4.7.1
- Added support for video and carousel ads inside MediaView


aserrr:
Conflicts resolved

PR #40, opened 2018-01-17
Adding DuAdPlatform adapter files

baiduDuApps:


adapters/DuAd/DuAdAdapter/GADDuAdNativeAd.m
```diff
@@ -0,0 +1,354 @@
...
+    [strongConnector adapter:strongAdapter didReceiveMediatedNativeAd:self];
+}
+
+#pragma mark - GADMediatedNativeAd
+
+- (id<GADMediatedNativeAdDelegate>)mediatedNativeAdDelegate {
+    return self;
+}
+
+#pragma mark - GADMediatedNativeAppInstallAd
```
sbagadi: Media view is now a required asset for native ads. If media view is not supported, you can return an `UIImageView` that shows the main image in it.

Please see this MoPub adapter implementation for more detail: https://github.com/googleads/googleads-mobile-ios-mediation/blob/6ed94c724a6e8ebf42219cbeb578292844669101/adapters/MoPub/MoPubAdapter/MoPubAdapterMediatedNativeAd.m#L197

Also, please make sure to always download the main image even if the `disableImageLoading` option is set to `YES` so that it can be shown in the media view.

adapters/DuAd/DuAdAdapter/GADDuAdNativeAd.m
```diff
@@ -0,0 +1,354 @@
...
+}
+
+- (void)nativeAd:(DUNativeAd *)nativeAd didFailWithError:(NSError *)error {
+    id<GADMAdNetworkConnector> strongConnector = _connector;
+    id<GADMAdNetworkAdapter> strongAdapter = _adapter;
+    [strongConnector adapter:strongAdapter didFailAd:error];
+}
+
+- (void)nativeAdDidClick:(DUNativeAd *)nativeAd {
+    [GADMediatedNativeAdNotificationSource mediatedNativeAdDidRecordClick:self];
```
sbagadi: Since a click would lead to an user leaving the application, you can send the `adapterWillLeaveApplication` callback here as.

baiduDuApps: I use StoreKit to show the AD details in apps, in this case, should I send the adapterWillLeaveApplication callback? (interstitial AD in the same situation)

sbagadi:
We are merging the PR as-is although some changes need to be made. We'll make those changes on our end and will update you when we have the updates ready.

Issue #50, opened 2018-02-19
Crash in [NSCache setObject:forKey:cost:] - attempt to insert nil value


Neral:
We noticed an increase in crashes while trying to load an ad.

It seems the issue is here: [InMobiMediatedNativeAppInstallAd.m#L86](https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/adapters/InMobi/InMobiAdapter/InMobiMediatedNativeAppInstallAd.m#L86)
If `NSData` fails to load an image it still going to set a nil object.

The issue introduced with commit https://github.com/googleads/googleads-mobile-ios-mediation/commit/1a952d127f9a23bbc00c691e25cf5722195d746c as the previous implementation has check for nil object https://github.com/googleads/googleads-mobile-ios-mediation/commit/1a952d127f9a23bbc00c691e25cf5722195d746c?diff=split#diff-0144ea4814b9001d3fffcd88ee007f84L112

### Crash
Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: '-[NSCache setObject:forKey:cost:]: attempt to insert nil value

### Debug log
```
Fatal Exception: NSInvalidArgumentException
-[InMobiMediatedNativeAppInstallAd initWithInMobiNativeAppInstallAd:withAdapter:
shouldDownloadImage:withCache:]_block_invoke
```

maninomula:
The PR has been merged and a new version of the adapter is release with the changes.

PR #51, opened 2018-02-19
Simplify InMobiMediatedNativeAppInstallAd init

Neral:
Updated `InMobiMediatedNativeAppInstallAd` initialization method.

### Tasks
- Simplified `init` method by extracting into few methods.
- Added additional check to prevent this crash: https://github.com/googleads/googleads-mobile-ios-mediation/issues/50
- Ensured that callback is in the main thread.
- Fallback to default Xcode code formation


Neral:
The wrong account used here.

PR #52, opened 2018-02-20
Simplify InMobiMediatedNativeAppInstallAd init

Neral:
Updated `InMobiMediatedNativeAppInstallAd` initialization method.

### Tasks
- Simplified `init` method by extracting into few methods.
- Added additional check to prevent this crash: https://github.com/googleads/googleads-mobile-ios-mediation/issues/50
- Ensured that callback is in the main thread.
- Fallback to default Xcode code formation

### DIFF
--- a/adapters/InMobi/InMobiAdapter/InMobiMediatedNativeAppInstallAd.m
+++ b/adapters/InMobi/InMobiAdapter/InMobiMediatedNativeAppInstallAd.m
@@ -99,8 +99,8 @@ static CGFloat const DefaultIconScale = 1.0;
             NSData *imageData = [NSData dataWithContentsOfURL:url];
             UIImage *image = [UIImage imageWithData:imageData];
             if (image) {
-                [imageCache setObject:cachedImage forKey:cacheKey];
                 cachedImage = image;
+                [imageCache setObject:cachedImage forKey:cacheKey];
             }
         }
         dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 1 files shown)


adapters/InMobi/InMobiAdapter/InMobiMediatedNativeAppInstallAd.m
```diff
@@ -23,100 +24,140 @@ @interface InMobiMediatedNativeAppInstallAd () <GADMediatedNativeAdDelegate,
...
-              [self inmobiMediatedNativeAppInstallAdFailed];
+- (void)loadImageWithURL:(NSURL *)url imageCache:(NSCache *)imageCache callback:(void(^)(UIImage *))callback {
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{
+        NSString *cacheKey = [url.absoluteString stringByAddingPercentEscapesUsingEncoding:NSUTF8StringEncoding];
+        UIImage *cachedImage = [imageCache objectForKey:cacheKey];
+        if (!cachedImage) {
+            NSData *imageData = [NSData dataWithContentsOfURL:url];
+            UIImage *image = [UIImage imageWithData:imageData];
+            if (image) {
+                [imageCache setObject:cachedImage forKey:cacheKey];
```
sbagadi: Here, the `cachedImage` is being added to the cache before being set to the `image` and is causing a crash since `cachedImage` is `nil`. The line below and this should be swapped to fix this.

Neral:
@sbagadi  Good catch. Updated.

PR #74, opened 2018-05-25
Update AdColony to 3.3.4 and forward consent for GDPR

eddieh:


adapters/AdColony/Public/Headers/GADMAdapterAdColonyExtras.h
```diff
@@ -4,7 +4,7 @@
 
 @import GoogleMobileAds;
 
-@interface GADMAdapterAdColonyExtras : NSObject<GADAdNetworkExtras>
+@interface GADMAdapterAdColonyExtras : GADExtras
```
maninomula: Please use GADNetworkExtras to pass the network specific extras. Using additionalParameters is not preferable.

adapters/AdColony/AdColonyAdapter/GADMAdapterAdColony.m
```diff
@@ -21,6 +21,9 @@
   INIT_STATE_INITIALIZING
 } InitState;
 
+static NSString *const kAdColonyExplicitConsentGiven = @"explicit_consent_given";
+static NSString *const kAdColonyConsentResponse = @"consent_response";
```
maninomula: Please use the methods mentioned by AdColony in [https://github.com/AdColony/AdColony-iOS-SDK-3/wiki/GDPR](https://github.com/AdColony/AdColony-iOS-SDK-3/wiki/GDPR) for passing consent values.

maninomula: Please use the methods mentioned by AdColony in [https://github.com/AdColony/AdColony-iOS-SDK-3/wiki/GDPR](https://github.com/AdColony/AdColony-iOS-SDK-3/wiki/GDPR) to pass the consent value.

adapters/AdColony/Public/Headers/GADMAdapterAdColonyExtras.h
eddieh: `additionalParameters` is where the `npa` value is stored. If there's another way to access that value I'd gladly use it. Is there another way?

adapters/AdColony/AdColonyAdapter/GADMAdapterAdColony.m
eddieh: These are the keys we're using at AdColony for forwarding consent from mediation. They're not publicly documented, but these are indeed the keys we want and that the ad server expects.

adapters/AdColony/Public/Headers/GADMAdapterAdColonyExtras.h
eddieh: Google's documentation says it will be in the `additionalParameters` so no matter what I have to access it through that. https://developers.google.com/admob/ios/eu-consent#forward_consent_to_the_google_mobile_ads_sdk

I can make a property for it, rather than accessing it directly.

maninomula: The documentation you provided is for publishers who are using Google Mobile Ads SDK only. For mediation we need to pass the required consent values through adapter specific extras class, which here is GADMAdapterAdColonyExtras class. Please create a static method in GADMAdapterAdColonyExtras to pass the consent values by the publisher.

eddieh:
We were under the impression that consent could be forwarded from the publisher's app → the mediation SDK → AdColony. I've updated this PR so that the publisher must forward consent to AdColony explicitly through the publicly document properties. Sorry about the confusion.

maninomula:
No problem. The updated code looks good to me.

PR #76, opened 2018-05-28
Use #import instead of @import

timlenardo:
This is a Unity specific nit. 

If you're building an iOS Unity project with this adapter included, the default behavior is to building without support for '@import' ('Enable Modules' setting in Build Settings). So the default build won't compile until you make the change I'm proposing, or enable modules in settings. Not a huge deal, but unless there's a reason to use '@import' instead of '#import', this cleans up the experience of using this adapter in Unity!


maninomula:
Thanks for the pull request.

PR #83, opened 2018-08-23
kAdIconImageKey --> kPrivacyIconImageName

chauduyphanvu1:
This is to fix the MoPub privacy info icon image not shown due to wrong attribute name.


maninomula:
This is already added in the 5.3.0.1 version which will be releasing in next few days.

PR #90, opened 2018-11-10
Compatible with Vungle SDK 6.3.2

ashinagawa:
- Compatible with Vungle SDK 6.3.2
- Removed consent message version


adapters/Vungle/VungleAdapter/VungleRouter.m
```diff
@@ -215,8 +215,7 @@ - (void)vungleAdPlayabilityUpdate:(BOOL)isAdPlayable
 - (void)vungleSDKDidInitialize {
   VungleSDK *sdk = [VungleSDK sharedSDK];
   if ([VungleRouterConsent getConsentStatus] > 0) {
-    [sdk updateConsentStatus:[VungleRouterConsent getConsentStatus]
-       consentMessageVersion:[VungleRouterConsent getConsentMessageVersion]];
+    [sdk updateConsentStatus:[VungleRouterConsent getConsentStatus] consentMessageVersion:@""];
```
maninomula: Why are we sending an empty string as the consentMessageVersion? Is it not needed for the publisher to set this value?

adapters/Vungle/VungleAdapter/VungleRouterConsent.m
```diff
@@ -10,23 +10,20 @@
...
-+ (void)updateConsentStatus:(VungleConsentStatus)status consentMessageVersion:(NSString *)version {
-  currentConsentStatus = status;
-  currentConsentMessageVersion = version;
++ (BOOL)updateConsentStatus:(VungleConsentStatus)consentStatus {
   VungleSDK *sdk = [VungleSDK sharedSDK];
-  if (sdk.initialized && status > 0) {
-    [sdk updateConsentStatus:status consentMessageVersion:version];
+  if (consentStatus > 0) {
+    currentConsentStatus = consentStatus;
+    [sdk updateConsentStatus:consentStatus consentMessageVersion:@""];
```
maninomula: Why are we sending an empty string as the consentMessageVersion? Is it not needed for the publisher to set this value?

adapters/Vungle/VungleAdapter/VungleRouter.m
ashinagawa: @maninomula Thanks for your feedback. Since Vungle uses `consentMessageVersion` for only native integration case, we decided to hide in mediation adapter level.

adapters/Vungle/VungleAdapter/VungleRouterConsent.m
ashinagawa: @maninomula The same as above. Since Vungle uses `consentMessageVersion` for only native integration case, we decided to hide in mediation adapter level.

PR #102, opened 2019-02-26
Adding Oath (Verizon Ads SDK) adapter v1.0.2.0

mmullin:
This pull request adds the new Inline and Interstitial adapters for the Verizon Ads SDK. We plan to add support for Native and Rewarded Video mediation to the Verizon Ads SDK in a future adapter update.

PR #103, opened 2019-02-27
Adding Oath (Verizon Ads SDK) Adapter v1.0.3.0

mmullin:
This pull request replaces #102. Please note that the adapter must be used with Verizon Ads SDK v1.0.3.

### DIFF
--- a/adapters/Oath/OathAdapter/GADMVASAdapterBaseClass.m
+++ b/adapters/Oath/OathAdapter/GADMVASAdapterBaseClass.m
@@ -8,14 +8,8 @@
 
 @protocol GADMAdNetworkAdapter;
 @protocol GADMAdNetworkConnector;
-@protocol GADMRewardBasedVideoAdNetworkAdapter;
 
-#define MM_SMARTBANNER_PHONE CGSizeMake(320, 50)
-#define MM_SMARTBANNER_TABLET CGSizeMake(728, 90)
-#define MM_SMARTBANNER_MED_CUTOFF 450
-#define MM_SMARTBANNER_MAX_CUTOFF 740
-
-static NSString * const kVASAdapterVersion = @"1.0.3.0";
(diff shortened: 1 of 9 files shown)


adapters/Oath/OathAdapter/GADMVASAdapterBaseClass.m
```diff
@@ -0,0 +1,365 @@
...
+- (CGRect)makeBannerRect:(GADAdSize)adSize
+{
+    // Consideration for smart banners ( http://bit.ly/1wAWn0r )
+    CGSize windowSize = [[UIApplication sharedApplication] keyWindow].bounds.size;
+    
+    if ( GADAdSizeEqualToSize(kGADAdSizeSmartBannerPortrait, adSize) ||
+        GADAdSizeEqualToSize(kGADAdSizeSmartBannerLandscape, adSize) ) {
+        int offsetX = (int)(windowSize.width / 2);
+        
+        if ( windowSize.height < MM_SMARTBANNER_MED_CUTOFF ) {
```
joshuajplagonera: Should these be checking for width instead of height?

When making a SmartBanner request for a phone, it's trying to return a size applicable for a tablet (MM_SMARTBANNER_TABLET) instead of for a phone, causing the returning banner to be too large for the ad view.

adapters/Oath/OathAdapter/GADMVASAdapterBaseClass.m
```diff
@@ -0,0 +1,365 @@
...
+            // MYDAS will attempt to return a 320x50, even when hsht < 320 / hswd < 50
+            return CGRectMake(0,0,-1,-1);
```
joshuajplagonera: Should this return MM_SMARTBANNER_PHONE size rather than -1, since the ad request will fail if this method returns a width less than 0?

Issue #105, opened 2019-03-14
Nend not not with SmartBanner


paynechu:
Nend's Banner Adapter will not work if using SmartBanner as the host. Even the SmartBanner's size is enough big.
That because of the code below. Other adapters don't have this problem.

```
- (void)getBannerWithSize:(GADAdSize)adSize {
  id<GADMAdNetworkConnector> strongConnector = self.connector;
  if (!GADAdSizeEqualToSize(adSize, kGADAdSizeBanner) &&           // 320x50
      !GADAdSizeEqualToSize(adSize, kGADAdSizeLargeBanner) &&      // 320x100
      !GADAdSizeEqualToSize(adSize, kGADAdSizeMediumRectangle) &&  // 300x250
      !GADAdSizeEqualToSize(adSize, kGADAdSizeLeaderboard)) {      // 728x90
    [strongConnector adapter:self didFailAd:nil];
    return;
  }
```

maninomula:
Please use the latest Nend Adapter version (5.1.1.0) and let us know if you have any more questions.

maninomula:
closing this issue as there is no activity for more than a month.

PR #109, opened 2019-04-04
Update banner size check for nend adapter

fan-c-yu:


fan-c-yu:
I signed it!

stowy:
@fan-c-yu can you explain the rationale behind this change?

fan-c-yu:
@stowy This is for smart banner. Ad which is smaller than placement size can be displayed.
#105

fan-c-yu:
We are rethinking about the change, withdraw once.

PR #118, opened 2019-04-23
Updating AdColony adapter to fix a crash in case of failed to fetch rewarded ad.

maninomula:
Updating AdColony adapter to fix a crash in case of failed to fetch rewarded ad.


houmie:
So is this now merged or is CLA still missing?

PR #119, opened 2019-05-16
Adapter 3.1 update

fritzhuie:
Changes:
- Minor lifecycle improvements
- Load API functionality added via metadata API
- Improved diagnostics

Notes:
Load API will prevent rewarded and interstitial placements from loading in UnityAds until "load" is called.

To enable load API: enable_metadata_load must be set, and the game ID must be whitelisted by UnityAds.

If load API is not enabled, metadata calls should not affect on the adapter's behavior.


ericleich:
Overriding CLA status as alice has an explicit commit saying signed CLA, and we have a separate email chain with Unity asking for this to be submitted.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.h
```diff
@@ -29,19 +29,17 @@
 - (void)initializeWithGameID:(NSString *)gameID;
 
 /// Requests a reward-based video ad with |adapterDelegate|.
-- (void)requestRewardedAdWithDelegate:
-    (id<GADMAdapterUnityDataProvider, UnityAdsExtendedDelegate>)adapterDelegate;
+- (void)requestRewardedAdWithDelegate:(NSString *)gameID
```
ericleich: Param doesn't match the name. Method name is requestRewardedAdWithDelegate:delegate: which is odd.

Instead of adding another param like this, you could instead add a getGameID method to the GADMAdapterUnityDataProvider protocol. This delegate object getting passed in is already providing placement ID via this data provider protocol.

ericleich:
Closing this PR, as it has been merged via https://github.com/googleads/googleads-mobile-ios-mediation/pull/122

PR #128, opened 2019-07-09
Added support for Chartboost SDK 8.0.0, banner ads

daniel-barros:
Support for Chartboost banner ads.


adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
```diff
@@ -35,6 +35,9 @@ @interface GADMAdapterChartboostSingleton () <ChartboostDelegate> {
   NSMutableArray<ChartboostInitCompletionHandler> *_completionHandlers;
 }
 
+@property (nonatomic) NSMapTable<CHBBanner *, id<CHBBannerDelegate>> *bannersToDelegates;
+@property (nonatomic) NSMutableArray<CHBBanner *> *loadingBanners;
```
joshuajplagonera: Does Chartboost support loading multiple banner ads using the same location?

If so, then I think we can remove this, since I don't see this array getting checked anywhere else.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
```diff
@@ -203,6 +208,26 @@ - (void)presentInterstitialAdForDelegate:
   [Chartboost showInterstitial:[adapterDelegate getAdLocation]];
 }
 
+#pragma mark - Banner methods
+
+- (void)configureBannerWithSize:(GADAdSize)adSize
```
joshuajplagonera: Can we remove the Banner implementation out of the Singleton class and put it in GADMAdapterChartboost?

Unlike the Interstitial and Rewarded implementation, Banners doesn't seem to follow a singleton pattern.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
```diff
@@ -203,6 +208,26 @@ - (void)presentInterstitialAdForDelegate:
...
+    if (extras.frameworkVersion && extras.framework) {
+        [Chartboost setFramework:extras.framework
+                     withVersion:extras.frameworkVersion];
+    }
+    
+    CHBBanner *banner = [[CHBBanner alloc] initWithSize:adSize.size location:location delegate:self];
+    banner.automaticallyRefreshesContent = NO;
+    [self.loadingBanners addObject:banner];
+    [self.bannersToDelegates setObject:delegate forKey:banner];
+    [banner showFromViewController:viewController];
```
joshuajplagonera: Is it necessary to call this before the Banner Ad has been loaded?

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
daniel-barros: We do support multiple banners with the same location.

This array is still needed though, please check out the didCacheAd:error: implementation (it was moved to GADMAdapterChartboostBanner.m).
We use it to keep the Chartboost banner objects in memory while they are being loaded.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
daniel-barros: That's a good point. I moved all the management of banners to a new class GADMAdapterChartboostBanner.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
daniel-barros: I think so. 

Chartboost banners can be cached (loaded) and shown. 
The first method just downloads the required data without actually displaying the ad. The second one, showFromViewController:, is the counterpart to GADBannerView loadRequest:.

Here Google's SDK is requesting that the adapter loads a banner.

daniel-barros:
Hey @joshuajplagonera, thanks for taking time for the review!
I responded your comments and refactored the banner code into a new class. Please let me know of any other concerns and follow up on the existing comments if you don't feel they were properly solved.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
joshuajplagonera: Yes, -loadRequest: is when the Ad Request is sent using the Google Mobile Ads SDK.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboost.m
```diff
@@ -104,10 +116,20 @@ - (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewCo
...
-  // Chartboost doesn't support banner ads.
-  NSError *error = GADChartboostErrorWithDescription(@"Chartboost Ads doesn't support banner ads.");
-  [_interstitialConnector adapter:self didFailAd:error];
+- (void)getBannerWithSize:(GADAdSize)adSize
+{
+    __weak typeof(self) weakSelf = self;
+    [self initializeChartboost:^(BOOL success) {
+        if (success) {
+            GADMAdapterChartboost *strongSelf = weakSelf;
+            UIViewController *viewController = [strongSelf->_interstitialConnector viewControllerForPresentingModalView];
```
joshuajplagonera: I think we can rename `_interstitialConnector` to just `_connector` since it's being used for Banners now.

joshuajplagonera: I think we can rename this from `_interstitialConnector` to just `_connector` since it's being used for Banners now.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostBanner.m
```diff
@@ -0,0 +1,85 @@
...
+#import "GADMAdapterChartboostBanner.h"
+
+@interface GADMAdapterChartboostBanner () <CHBBannerDelegate>
+
+@property (nonatomic) NSMapTable<CHBBanner *, id<CHBBannerDelegate>> *bannersToDelegates;
+@property (nonatomic) NSMutableArray<CHBBanner *> *loadingBanners;
+
+@end
+
+@implementation GADMAdapterChartboostBanner
```
joshuajplagonera: I believe this still uses a Singleton pattern. The reason why we created a Singleton class for Interstitial and Rewarded is to handle instances because AdMob allows multiple Interstitials/Rewarded Ads instances while Chartboost only allows one (has a singleton pattern).

That said, is it possible to move the Banner implementation to the main adapter class (GADMAdapterChartboost) since Chartboost allows multiple Banner instances? It also removes the need to forward events from this class to the main adapter.

Let us know if there is a specific reason why a Singleton would be preferred for the implementation of Chartboost Banner ads.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboost.m
daniel-barros: That makes sense. Done

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostBanner.m
daniel-barros: This makes sense now, I had misunderstood how the GADMAdNetworkAdapter works. Thanks for the explanation!
I moved the banner implementation to the GADMAdapterChartboost class as you suggested.

daniel-barros:
Hi @joshuajplagonera, please let me know if there is anything we can do to move this PR forward (besides the CLA).

daniel-barros:
Hi @joshuajplagonera! Please let us know if there's something we can do to move forward the approval.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboost.m
```diff
@@ -159,15 +188,61 @@ - (void)didFailToLoadInterstitial:(CBLocation)location withError:(CBLoadError)er
...
+- (void)didCacheAd:(CHBCacheEvent *)event error:(nullable CHBCacheError *)error
+{
+    id<GADMAdNetworkConnector> strongConnector = _connector;
+    if (error) {
+        [strongConnector adapter:self didFailAd:NSErrorWithCHBCacheError(error)];
+    } else {
+        [strongConnector adapter:self didReceiveAdView:(UIView *)event.ad];
+    }
+    // We keep a strong reference to the banner only until it is loaded, since at that point the view is sent to GMA and it is its responsibility to retain it.
+    _loadingBanner = nil;
```
maninomula: Does nilling here also removes the banner delegate? Do we receive ad events even after nilling out? Because in line 134 we added this class as a delegate.

daniel-barros: It doesn't remove the banner delegate because at this point some other object must have kept a strong reference to the banner. The superview it has been place inside at least.
We receive ad events after nilling it out. The banner delegate is the adapter itself, which should be kept alive, and the banner itself isn't released.
The point in doing this is to keep a reference to the banner only as long as it is necessary to prevent memory leaks.

PR #139, opened 2019-08-12
Update Facebook Adapter to version 5.5.0.0

xilosada:
Prepare adapters for Facebook release on August 19.


maninomula:
We had made an internal PR and the Facebook adapter version 5.5.0.0 is live.

PR #141, opened 2019-08-26
Fix issue where UnityAds onReady callback fails to send isReady callback

fritzhuie:
**Issue**
The `unityAdsReady:` method in `GADMAdapterUnitySingleton` does not set the callback delegate properly, so the first load request does not inform the Admob SDK that it's ready

**Change**
Set the delegate, then send the callback before removing the delegate from the delegate map


maninomula:
Added this change internally.

PR #149, opened 2019-09-17
* Updating AppLovin adapter to be compatible with AppLovin SDK version 6.8.1.

maninomula:
* Updating AppLovin adapter to be compatible with AppLovin SDK version 6.8.1.
* Added code to properly handle the completion handlers after called.
* Linting the code to follow Google's Objective-C code-style.

### DIFF
--- a/adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovin.m
+++ b/adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovin.m
@@ -7,403 +7,184 @@
 //
 
 #import "GADMAdapterAppLovin.h"
-#import <AppLovinSDK/AppLovinSDK.h>
+#import "GADMAdapterAppLovinBannerDelegate.h"
 #import "GADMAdapterAppLovinConstant.h"
 #import "GADMAdapterAppLovinExtras.h"
+#import "GADMAdapterAppLovinInterstitialDelegate.h"
+#import "GADMAdapterAppLovinMediationManager.h"
 #import "GADMAdapterAppLovinUtils.h"
 
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
 
-/// Banner Delegate.
-@interface GADMAdapterAppLovinBannerDelegate
-    : NSObject <ALAdLoadDelegate, ALAdDisplayDelegate, ALAdViewEventDelegate>
-- (instancetype)initWithParentAdapter:(GADMAdapterAppLovin *)parentAdapter;
-@property(nonatomic, weak) GADMAdapterAppLovin *parentAdapter;
-@end
+@implementation GADMAdapterAppLovin {
+  /// Instance of the AppLovin SDK.
+  ALSdk *_sdk;
 
-/// AppLovin Interstitial Delegate.
-@interface GADMAdapterAppLovinInterstitialDelegate
-    : NSObject <ALAdLoadDelegate, ALAdDisplayDelegate, ALAdVideoPlaybackDelegate>
-- (instancetype)initWithParentRenderer:(GADMAdapterAppLovin *)parentRenderer;
-@property(nonatomic, weak) GADMAdapterAppLovin *parentRenderer;
-@end
+  /// AppLovin interstitial object used to request an ad.
+  ALInterstitialAd *_interstitial;
 
-@interface GADMAdapterAppLovin ()
+  /// AppLovin interstitial delegate wrapper.
+  GADMAdapterAppLovinInterstitialDelegate *_interstitialDelegate;
 
-/// Controlled Properties.
-@property(nonatomic, strong) ALSdk *sdk;
-@property(nonatomic, weak) id<GADMAdNetworkConnector> connector;
-
-/// Interstitial Properties.
-@property(nonatomic, strong, nullable) ALInterstitialAd *interstitial;
-@property(nonatomic, strong, nullable) ALAd *alInterstitialAd;
-@property(nonatomic, strong) GADMAdapterAppLovinInterstitialDelegate *interstitialDelegate;
-/// Banner Properties.
-@property(nonatomic, strong, nullable) ALAdView *adView;
-@property(nonatomic, strong) GADMAdapterAppLovinBannerDelegate *bannerDelegate;
-
-/// Controller properties - The connector/credentials referencing these properties may get
-/// deallocated.
-@property(nonatomic, copy) NSString *zoneIdentifier;
-
-@property(nonatomic) GADAdSize adSize;
-@property(nonatomic, strong) ALAd *ad;
-@end
-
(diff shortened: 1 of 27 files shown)


maninomula:
Changed the CLA as the changes are made by me.

adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovin.m
```diff
@@ -15,296 +15,298 @@
...
-  }
+  });
 
-  if ([requestedZones containsObject:self.zoneIdentifier]) {
-    [GADMAdapterAppLovinUtils log:@"Can't request a second ad for the same zone Identifier "
-                                  @"without showing the first ad."];
-    NSError *error = [NSError errorWithDomain:GADMAdapterAppLovinConstant.errorDomain
-                                         code:0
-                                     userInfo:nil];
+  if ([requestedZones containsObject:_zoneIdentifier]) {
```
thomasmso: If a publishers has 1 AdMob ad unit set up, with only 1 AppLovin zone, if AdMob times out the AppLovin ad load, this will disqualify any future AppLovin ad load attempts... even if we successfully returned an ad in `adService:didLoadAd:`.

I understand that the goal of this logic is to prevent loading two AppLovin ads of the same zone for pubs who have back-to-back ads. Let me check internally to see if we can turn on a feature to allow back-to-back ads.

maninomula: If it times out we are removing that particular zone identifier at the dealloc time on line 108, which doesn't disqualify future requests.

thomasmso: Ok there is a race condition there then, I believe [`dispatch_async(...`](https://github.com/googleads/googleads-mobile-ios-mediation/pull/149/files/5ec7d3f81b4564ba0b21b2f1eac8a4d1b2e60c7b#diff-78899a941343e15821a32b4276843661R106) should be `dispatch_sync`... Otherwise, `_interstitial = nil;` gets called first, causing the `if (self->_interstitial) {...}` check to always fail and zone to never be removed from the set.

maninomula: Added code to properly handle this case.

maninomula:
@thomasmso Added code to handle the race condition properly. Can you please take a look again?

thomasmso:
Yes, I checked out the piece of code again and it looks like the race condition is no longer there. Thank you!

maninomula:
@thomasmso  If everything looks good can you please approve the PR?

ericleich:
@maninomula to update this PR with changes since our internal code review since @thomasmso reviewed, so that @thomasmso can double check the changes. 

@thomasmso note that we have also recently published https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/CONTRIBUTING.md#best-practices outlining some practices that were applied in this PR, and that we will try to maintain going forward.

thomasmso:
@ericleich - got it. Also, our latest SDK (9.9.2) was released end of last week, no API / functionality changes.

maninomula:
I have updated the PR with all the internal changes. @thomasmso Please take a look.

adapters/AppLovin/AppLovinAdapter/GADMediationAdapterAppLovin.m
```diff
@@ -20,53 +20,65 @@
...
+      GADMAdapterAppLovinMutableSetAddObject(sdkKeys, sdkKey);
+    }
   }
 
-  // Log to the publisher that having more than 1 SDK key is... abnormal
-  if (sdkKeys.count > 1) {
-    [GADMAdapterAppLovinUtils
-        log:@"Found %lu SDK keys. Please remove any SDK keys you are not using from the AdMob UI.",
-            sdkKeys.count];
+  // Log to the publisher that having more than 1 SDK key is... abnormal.
```
thomasmso: Should it be `sdkKeys.count != 1`? Or is only the error case when there are no SDK keys provided? If so, comment does not reflect the error case.

adapters/AppLovin/AppLovinAdapter/GADMediationAdapterAppLovin.m
```diff
@@ -20,53 +20,65 @@
...
+          (unsigned long)sdkKeys.count];
+
+  // Initialize SDKs based on SDK keys.
   NSSet<NSString *> *sdkKeysCopy = [sdkKeys copy];
   for (NSString *sdkKey in sdkKeysCopy) {
     [GADMAdapterAppLovinUtils log:@"Initializing SDK for SDK key %@", sdkKey];
 
     ALSdk *sdk = [GADMAdapterAppLovinUtils retrieveSDKFromSDKKey:sdkKey];
     [sdk initializeSdkWithCompletionHandler:^(ALSdkConfiguration *configuration) {
       @synchronized(sdkKeysLock) {
```
thomasmso: We don't need this `@synchronized` block anymore if we're using the new atomic functions now.

adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovinConstant.m
```diff
@@ -8,34 +8,18 @@
...
 
-+ (NSString *)rtbErrorDomain {
-  return @"com.applovin.sdk.mediation.admob.rtb.errorDomain";
-}
+NSString *const GADMAdapterAppLovinInfoPListSDKKey = @"AppLovinSdkKey";
 
-+ (NSString *)adapterVersion {
-  return @"6.8.0.0";
-}
+NSString *const GADMAdapterAppLovinAdapterVersion = @"6.8.0.0";
```
thomasmso:
```suggestion
NSString *const GADMAdapterAppLovinAdapterVersion = @"6.9.4.0";
```

adapters/AppLovin/AppLovinAdapter/GADMediationAdapterAppLovin.m
maninomula: removed

adapters/AppLovin/AppLovinAdapter/GADMediationAdapterAppLovin.m
maninomula: We are only throwing an error only if the `sdkKeys` set is empty, so the comment is no longer correct here, removed.

adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovinConstant.m
maninomula: This will be done in another PR.

maninomula:
@thomasmso Apart from the version number update which will be done in the next PR, if everything looks okay please approve the changes. Thanks !

PR #154, opened 2019-09-24
Releasing AdColony Adapter v4.1.2.0

mahesh-kokate:


maninomula:
Hello @mahesh-kokate, we have updated and released the AdColony adapter to version 3.3.8.1.0. Now will be starting working on the 4.1.1.0 version. Can you please resolve the conflicts so that we can move forward?

mahesh-kokate:
Sure @maninomula, I'll work on it.

mahesh-kokate:
Hi @maninomula,

We need help regarding notifying ad expire event to a publisher.

As you know, currently we are just printing logs from the ad expire block in the AdColony adapter, there is no way for a publisher to know about this event. So can you advise us how we can forward this event to a publisher/app? 

There are some cases where a publisher needs this event. 

For example:
1 - App initialized AdMob SDK by using App ID associated with AdColony (Which will also initialize/configure AdColony SDK)
2 - After completion of initialization, sends an ad request (consider a case where it will return successful loading callback).
3 - Then after some time suppose 15-20 mins if the app tries to show the ad using the above ad object, it fails to present it since it is expired already. 

In this case, there is no way for the app to know that it needs to make a new fresh ad request instead of using a stale ad object.

Earlier we were re-requesting ad request internally from ad expire block but this will not work for open bidding logic.

We also tried by calling the failure delegate callback from AdColony adapter's ad expire block, but it's not going to the app. It might be because of waterfall logic inside the AdMob SDK.

So please advise us to handle the ad expire callback.

Thanks,
Mahesh Kokate

maninomula:
@mahesh-kokate, Thanks for brining up the issue. As discussed we have added code to re-request ads for OpenBidding requests if the loaded ad expires. There is a PR for this implementation. I am closing this PR for now as the code had already been submitted and adding you as a reviewer for that PR.

PR #160, opened 2019-10-09
Added code which prevents AppLovin adapter to request multiple ads using the same zone identifier.

maninomula:
Added code which prevents AppLovin adapter to request multiple ads using the same zone identifier.


maninomula:
@thomasmso, this PR has code which prevents AppLovin adapter from requesting multiple ads using the same zone identifier. So, can you please review this PR which unblocks us from releasing the latest version of the AppLovin adapter.

PR #169, opened 2019-11-14
Updated Facebook adapter to fail if the app or service is primarily child directed.

maninomula:
Updated Facebook adapter to fail if the app or service is primarily child directed.


maninomula:
I have made a PR, so manually changed the CLA.

PR #170, opened 2019-11-25
Update maio adapter v1.5.1.0

im-ttmskk:
- Update adapter version


maninomula:
Thanks for the PR. We have released this version of the Maio adapter. Closing because this is already live.

PR #178, opened 2019-12-13
Updated Facebook adapter to their latest API to pass the COPPA info.

maninomula:
Updated Facebook adapter to their latest API to pass the COPPA info.


maninomula:
Hello @yroxe and @xilosada, can either of you take a look at this PR so that we can merge this?

xilosada:
Hi @maninomula, FB code looks good to me but this PR includes changes for TapJoy. I think it will be better to split it in two branches.

maninomula:
@xilosada, thanks for reviewing. Removed Tapjoy code from this PR.

PR #185, opened 2020-01-29
Updating Fyber Adapter to version 7.5.1

AviGelkop:
Update the adapter to version 7.5.1
Fixed bug: calling `dispatch_sync(dispatch_get_main_queue()` caused crash. There is no need to do it on the main thread since the init flow will continue with another thread anyway.


AviGelkop:
@googlebot I signed it!

PR #186, opened 2020-01-29
Vungle iOS AdMob Adapter 6.5.1.1 Release

ashinagawa:
This build is compatible with Vungle iOS SDK 6.5.1.

### DIFF
--- a/adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
+++ b/adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
@@ -37,4 +37,6 @@
 
 @property(nonatomic, copy) NSString *_Nullable playingPlacement;
 
+@property(nonatomic, copy) NSNumber *_Nullable orientations;
+
 @end
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.4.6.0";
+static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.5.0.0";
 static NSString *const _Nonnull kGADMAdapterVungleApplicationID = @"application_id";
 static NSString *const _Nonnull kGADMAdapterVunglePlacementID = @"placementID";
 static NSString *const _Nonnull kGADMAdapterVungleErrorDomain = @"com.google.mediation.vungle";
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.5.0.0";
+static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.5.1.0";
 static NSString *const _Nonnull kGADMAdapterVungleApplicationID = @"application_id";
 static NSString *const _Nonnull kGADMAdapterVunglePlacementID = @"placementID";
 static NSString *const _Nonnull kGADMAdapterVungleErrorDomain = @"com.google.mediation.vungle";
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.5.1.0";
+static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.5.1.1";
 static NSString *const _Nonnull kGADMAdapterVungleApplicationID = @"application_id";
 static NSString *const _Nonnull kGADMAdapterVunglePlacementID = @"placementID";
 static NSString *const _Nonnull kGADMAdapterVungleErrorDomain = @"com.google.mediation.vungle";
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
@@ -14,10 +14,13 @@
 
 /// Vungle adapter ad type.
 typedef NS_ENUM(NSUInteger, GADMAdapterVungleAdType) {
-  GADMAdapterVungleAdTypeUnknown,       ///< Unknown adapter type.
-  GADMAdapterVungleAdTypeRewarded,      ///< Rewarded adapter type.
-  GADMAdapterVungleAdTypeInterstitial,  ///< Interstitial adapter type.
-  GADMAdapterVungleAdTypeBanner         ///< Banner adapter type.
+  GADMAdapterVungleAdTypeUnknown,            ///< Unknown adapter type.
+  GADMAdapterVungleAdTypeRewarded,           ///< Rewarded adapter type.
+  GADMAdapterVungleAdTypeInterstitial,       ///< Interstitial adapter type.
+  GADMAdapterVungleAdTypeMREC,               ///< MREC adapter type.
+  GADMAdapterVungleAdTypeBanner,             ///< Banner adapter type.
+  GADMAdapterVungleAdTypeShortBanner,        ///< ShortBanner adapter type.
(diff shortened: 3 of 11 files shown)


ashinagawa:
@googlebot I consent.

YueVungle:
@googlebot I consent.

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
```diff
@@ -37,4 +37,6 @@
 
 @property(nonatomic, copy) NSString *_Nullable playingPlacement;
 
+@property(nonatomic, copy) NSNumber *_Nullable orientations;
```
joshuajplagonera: How about let publishers pass a `UIInterfaceOrientationMask` here? I assume Vungle's SDK does the massaging of this parameter. API doc can say only `UIInterfaceOrientationLandscape`, `Portrait`, or `All` are supported.

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
```diff
@@ -14,10 +14,13 @@
...
-  GADMAdapterVungleAdTypeRewarded,      ///< Rewarded adapter type.
-  GADMAdapterVungleAdTypeInterstitial,  ///< Interstitial adapter type.
-  GADMAdapterVungleAdTypeBanner         ///< Banner adapter type.
+  GADMAdapterVungleAdTypeUnknown,            ///< Unknown adapter type.
+  GADMAdapterVungleAdTypeRewarded,           ///< Rewarded adapter type.
+  GADMAdapterVungleAdTypeInterstitial,       ///< Interstitial adapter type.
+  GADMAdapterVungleAdTypeMREC,               ///< MREC adapter type.
+  GADMAdapterVungleAdTypeBanner,             ///< Banner adapter type.
+  GADMAdapterVungleAdTypeShortBanner,        ///< ShortBanner adapter type.
+  GADMAdapterVungleAdTypeLeaderboardBanner   ///< LeaderboardBanner adapter type.
```
joshuajplagonera: Are these needed? Is it necessary to have 4 types for Banner ads? Can we stick to one Banner `GADMAdapterVungleAdType`?

adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.4.6.0";
+static NSString *const _Nonnull kGADMAdapterVungleVersion = @"6.5.2.0";
```
joshuajplagonera: This PR is for `6.5.1.1` release, as per the CHANGELOG and title. Is this correct?

adapters/Vungle/VungleAdapter/GADMAdapterVungleRewardedAd.m
```diff
@@ -35,6 +35,10 @@ @implementation GADMAdapterVungleRewardedAd {
   BOOL _isRewardedAdPresenting;
 }
 
++ (NSString *)adapterVersion {
+    return kGADMAdapterVungleVersion;
+}
```
joshuajplagonera: What is this for? It doesn't look like it is used anywhere?

Can we remove this and use the `+version:` from `GADMediationAdapterVungle`, or just use `kGADMAdapterVungleVersion` directly if this is needed?

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -186,7 +191,10 @@ - (BOOL)hasDelegateForPlacementID:(nonnull NSString *)placementID
 }
 
 - (BOOL)canRequestBannerAdForPlacementID:(nonnull NSString *)placmentID {
-  return _bannerPlacementID == nil || [_bannerPlacementID isEqualToString:placmentID];
+  if (_bannerDelegates.count > 0) {
+    return _bannerPlacementID == nil || [_bannerPlacementID isEqualToString:placmentID];
+    }
+  return YES;
```
joshuajplagonera: Can Vungle SDK now load multiple banners for different sizes? How about multiple banners for the same size? If there are no delegates, shouldn't `placementID` be `nil` and this should have passed anyways?

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -204,24 +212,46 @@ - (nullable NSError *)loadAd:(nonnull NSString *)placement
     [delegate adAvailable];
   } else {
     NSError *loadError;
-    if (![sdk loadPlacementWithID:placement error:&loadError]) {
-      if (loadError) {
-        return loadError;
-      }
+    GADMAdapterVungleAdType adType = [delegate adapterAdType];
+    if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner && adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
```
joshuajplagonera: So Vungle MREC ads are requested differently than the rest of banners? Maybe in that case we need 2 enum values, 1 for MREC and only one other for "banner"?

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
YueVungle: @joshuajplagonera  Hi joshuajplagonera, because Vungle SDK also uses NSNumber type for orientations options, here we NSNumber type is more convenient. Thanks a lot. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
YueVungle: @joshuajplagonera Hi joshuajplagonera, all of these types are needed. Because MREC ads and Banner ads are using different APIs to load ads. When loading Banner ads, we also need to pass the Banner size type (Banner[320 * 50], short Banner[300 * 50] or Leaderboard Banner[728*90]) to Vungle SDK,  so we need different Banner types to distinguish Banner ads of different sizes. Thanks a lot. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
YueVungle: @joshuajplagonera Hi joshuajplagonera, I'm really sorry for the confusing. Because I submitted a modification to support Adaptive and Smart Banners, which updated the kGADMAdapterVungleVersion from 6.5.1.1 to 6.5.2.0. I will update the CHANGELOG right now.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @joshuajplagonera Hi joshuajplagonera, we don't support multiple banners right now. If there are no delegates, it will return YES directly, not check placementID is nil(in this case, I think the placementID should be nil). Thanks a lot.  🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @joshuajplagonera Hi joshuajplagonera, yes, Vungle MREC ads and Vungle Banner ads are using different APIs to load ads. Same question as above, all of these Banner enums are needed. When loading Vungle Banner ads, we also need to pass the Banner size type (Banner[320 * 50], short Banner[300 * 50] or Leaderboard Banner[728*90]) to Vungle SDK, so we need different Banner enum types to distinguish Banner ads of different sizes. Thanks a lot. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleConstants.h
YueVungle: @joshuajplagonera Hi joshuajplagonera, I have updated the CHANGELOG for 6.5.2.0 version. I'm really sorry for the confusing and thanks a lot for the reminder. Have a nice day. 😊

YueVungle: @joshuajplagonera Hi joshuajplagonera, after we discussed, since we will release 6.5.2 SDK soon, we will hold this PR for Vungle AdMob Adapter 6.5.1.1. And we will create a new PR for Vungle AdMob Adapter 6.5.2 to support "Smart and Adaptive Banners". So I reset this PR to the previous 6.5.1.1 state. I'm really sorry for the confusing and have a nice day. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRewardedAd.m
YueVungle: @joshuajplagonera Hi joshuajplagonera, the class method `adapterVersion` is only used from our QA Test app to display adapter version number in log. So I removed this method from `GADMAdapterVungleInterstitial` and `GADMAdapterVungleRewardedAd` class. Thanks very much for your suggestions.🙂

PR #195, opened 2020-02-20
Unity3.4.2

jindoujiao:
Update to support UnityAds SDK 3.4.2


adapters/Unity/UnityAdapter/GADMAdapterUnity.m
```diff
@@ -81,17 +80,25 @@ - (void)getInterstitial {
...
+  UnitySingletonCompletion completeBlock = ^(UnityAdsError *error, NSString *message) {
+    if(!error) {
+            [UnityAds addDelegate:self];
+            [UnityAds load:[self getPlacementID]];
+    } else {
+      if (strongConnector) {
+        NSError *errorWithDescription = GADUnityErrorWithDescription(message);
+        [strongConnector adapter:self didFailAd:errorWithDescription];
+      }
+    }
```
joshuajplagonera: Can be simplified to avoid nesting? Apply through out all files.

```
if (error) {
  if (strongConnector) {
    NSError *errorWithDescription = GADUnityErrorWithDescription(message);
    [strongConnector adapter:self didFailAd:errorWithDescription];
  }
  return;
}

[UnityAds addDelegate:self];
[UnityAds load:_placementID];
```

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
```diff
@@ -123,68 +130,66 @@ - (NSString *)getPlacementID {
...
+    if (newState == kUnityAdsPlacementStateNoFill){
+      NSError *errorWithDescription = GADUnityErrorWithDescription(@"NO_FILL");
+      [strongNetworkConnector adapter:self didFailAd:errorWithDescription];
+    } else if (newState == kUnityAdsPlacementStateDisabled) {
+      NSError *errorWithDescription = GADUnityErrorWithDescription(@"PlACEMENT_DISABLED");
+      [strongNetworkConnector adapter:self didFailAd:errorWithDescription];
+    } else if (newState == kUnityAdsPlacementStateNotAvailable) {
+      NSError *errorWithDescription = GADUnityErrorWithDescription(@"PlACEMENT_NOTAVAILABLE");
+      [strongNetworkConnector adapter:self didFailAd:errorWithDescription];
+    }
```
joshuajplagonera: We can move `[strongNetworkConnector adapter:self didFailAd:errorWithDescription];` out of the `if()` and `else if()` blocks.

Additionally, can this be turned into a `switch()` statement with a `default` returning a generic error to capture all possible outcomes?

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
```diff
@@ -123,68 +130,66 @@ - (NSString *)getPlacementID {
...
-    }
-    return;
-  }
-
-  NSError *errorWithDescription = GADUnityErrorWithDescription(message);
-  if (strongNetworkConnector) {
-    [strongNetworkConnector adapter:self didFailAd:errorWithDescription];
-  }
-  _isLoading = NO;
+  //do nothing.
```
joshuajplagonera: Can you confirm that it is safer to rely on `unityAdsPlacementStateChanged` to catch errors vs catching them on `unityAdsDidError` ?

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -37,6 +34,8 @@ @interface GADMAdapterUnityRewardedAd () <GADMAdapterUnityDataProvider, UnityAds
   BOOL _isLoading;
 }
 
+// The completion handler to call when the ad loading succeeds or fails.
+@property (nonatomic, copy)GADMediationRewardedLoadCompletionHandler adLoadCompletionHandler;
```
joshuajplagonera: Is there a reason why this was changed to a property?

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -117,70 +112,73 @@ - (NSString *)getPlacementID {
...
+  }
 }
 
-- (void)unityAdsPlacementStateChanged:(nonnull NSString *)placementId
+- (void)unityAdsPlacementStateChanged:(nonnull NSString *)placementID
                              oldState:(UnityAdsPlacementState)oldState
                              newState:(UnityAdsPlacementState)newState {
   // This callback is not forwarded to the adapter by the GADMAdapterUnitySingleton and the adapter
   // should use the unityAdsReady: and unityAdsDidError: callbacks to forward Unity Ads SDK state to
   // Google Mobile Ads SDK.
```
joshuajplagonera: I believe we can remove this since this is no longer true.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -42,189 +33,62 @@ + (instancetype)sharedInstance {
...
-                                              valueOptions:NSMapTableWeakMemory];
-  }
   return self;
 }
 
-- (void)initializeWithGameID:(NSString *)gameID {
+- (void)initializeWithGameID:(NSString *)gameID
+                    completeBlock:(UnitySingletonCompletion)completeBlock{
   if ([UnityAds isInitialized]) {
+    completeBlock(NULL, @"UnityAds Initialization Succeeded");
```
joshuajplagonera: `NULL` -> `nil`
Please apply through out all files.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -42,189 +33,62 @@ + (instancetype)sharedInstance {
...
     return;
   }
 
-  // Metadata needed by Unity Ads SDK before initialization.
+  [self setCompleteBlock:completeBlock];
```
joshuajplagonera: Can we just do `self.completeBlock = completeBlock;` here?

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -42,189 +33,62 @@ + (instancetype)sharedInstance {
...
-  // This callback is not forwarded to the adapter by the GADMAdapterUnitySingleton and the adapter
-  // should use the unityAdsReady: and unityAdsDidError: callbacks to forward Unity Ads SDK state to
-  // Google Mobile Ads SDK.
+    if (newState == kUnityAdsPlacementStateWaiting || newState == kUnityAdsPlacementStateReady) {
+      if (self.completeBlock) {
+        self.completeBlock(NULL, @"UnityAds Initialization Succeeded");
+        self.completeBlock = nil;
+      }
+    }
+    return;
```
joshuajplagonera: I think we can remove the `return` here and below.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -42,189 +33,62 @@ + (instancetype)sharedInstance {
...
                              newState:(UnityAdsPlacementState)newState {
-  // This callback is not forwarded to the adapter by the GADMAdapterUnitySingleton and the adapter
-  // should use the unityAdsReady: and unityAdsDidError: callbacks to forward Unity Ads SDK state to
-  // Google Mobile Ads SDK.
+    if (newState == kUnityAdsPlacementStateWaiting || newState == kUnityAdsPlacementStateReady) {
+      if (self.completeBlock) {
+        self.completeBlock(NULL, @"UnityAds Initialization Succeeded");
+        self.completeBlock = nil;
+      }
+    }
```
joshuajplagonera: Combine into one `if()` block?

```
if ((newState == kUnityAdsPlacementStateWaiting || newState == kUnityAdsPlacementStateReady) && self.completeBlock) {
  self.completeBlock(nil, @"UnityAds Initialization Succeeded");
  self.completeBlock = nil;
}
```

adapters/Unity/UnityAdapter/GADMediationAdapterUnity.m
```diff
@@ -51,9 +51,17 @@ + (void)setUpWithConfiguration:(GADMediationServerConfiguration *)configuration
...
-
-  [[GADMAdapterUnitySingleton sharedInstance] initializeWithGameID:gameID];
-  completionHandler(nil);
+  
+  UnitySingletonCompletion completeBlock = ^(UnityAdsError *error, NSString *message) {
+    if(!error && completionHandler) {
+      completionHandler(nil);
+    } else {
+      completionHandler(GADUnityErrorWithDescription(message));
+    }
```
joshuajplagonera: Simplify?

```
if (error) {
  completionHandler(GADUnityErrorWithDescription(message));
  return;
}
completionHandler(nil);
```

jindoujiao:
@joshuajplagonera Thanks for the comments. I tried to address them and commit changes accordingly. Please review.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
jindoujiao: the intent is to access within the complete block.

joshuajplagonera: I believe we can create a strong reference to the adapter within the block in order to access `adLoadCompletionHandler` as an instance variable.

```
GADMAdapterUnityRewardedAd *__weak weakSelf = self;
UnitySingletonCompletion completeBlock = ^(UnityAdsError *error, NSString *message) {
  GADMAdapterUnityRewardedAd *strongSelf = weakSelf;
  if (!strongSelf) {
    return;
  }

  if (error) {
    NSError *errorWithDescription = GADUnityErrorWithDescription(message);
    strongSelf->_adLoadCompletionHandler(nil, errorWithDescription);
    return;
  }
  [UnityAds addDelegate:strongSelf];
  [UnityAds load:[strongSelf getPlacementID]];
};
```

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
joshuajplagonera: Thank you for clarifying.

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
```diff
@@ -81,17 +81,25 @@ - (void)getInterstitial {
...
+
+  UnitySingletonCompletion completeBlock = ^(UnityAdsError *error, NSString *message) {
+    if(error) {
+      if (strongConnector) {
+        NSError *errorWithDescription = GADUnityErrorWithDescription(message);
+        [strongConnector adapter:self didFailAd:errorWithDescription];
+      }
+      return;
+    }
+    [UnityAds addDelegate:self];
```
ericleich: Even with the perPlacement load API, UnityAds can still only load 1 ad per placement at a time, correct? This change no longer goes through the singleton to make sure another ad wasn't already loaded for that placement.

My guess is for each callback, UnityAds invokes every registered delegate with that callback. If 2 interstitial adapters are listening in on the same placement, then 2 adapters will claim an ad has loaded when there's only 1 backing ad. That is why we previously went through the singleton class, to make sure that only 1 adapter per placement would stick around.

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
```diff
@@ -123,68 +131,66 @@ - (NSString *)getPlacementID {
...
 }
 
 - (void)unityAdsReady:(NSString *)placementID {
   id<GADMAdNetworkConnector> strongNetworkConnector = _networkConnector;
-  if (!_isLoading) {
-    return;
-  }
-
-  if (strongNetworkConnector) {
+  if (strongNetworkConnector && [placementID isEqualToString:_placementID]) {
```
ericleich: What if the initialization call gets us into a READY state before this class gets the init complete callback and register this class as a delegate. Will UnityAds SDK still send the unityAdsReady callback to this object? It seems we risk a race condition where we could miss a READY call.

It would be helpful to understand possible state diagram with placementStateChanged. i.e. will we always go into kUnityAdsPlacementStateWaiting a run loop prior to ready?

adapters/Unity/UnityAdapter/GADMAdapterUnityBannerAd.m
```diff
@@ -57,7 +57,18 @@ - (void)loadBannerWithSize:(GADAdSize)adSize {
...
     NSString *gameID = [strongConnector.credentials[kGADMAdapterUnityGameID] copy];
-    [GADMAdapterUnitySingleton.sharedInstance initializeWithGameID:gameID];
+      
+    UnitySingletonCompletion completeBlock = ^(UnityAdsError *error, NSString *message) {
+        if(error) {
+          if (strongConnector && strongAdapter) {
+            [strongConnector adapter:strongAdapter didFailAd:GADUnityErrorWithDescription(message)];
+          }
+          return;
+        }
```
ericleich: Move the code that loads a banner into this completion block?

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -117,70 +115,69 @@ - (NSString *)getPlacementID {
...
+      [strongDelegate didEndVideo];
+      // Unity Ads doesn't provide a way to set the reward on their front-end. Default to a reward
+      // amount of 1. Publishers using this adapter should override the reward on the AdMob
+      // front-end.
+      GADAdReward *reward = [[GADAdReward alloc] initWithRewardType:@""
+                                                         rewardAmount:[NSDecimalNumber one]];
+      [strongDelegate didRewardUserWithReward:reward];
+    } else if (state == kUnityAdsFinishStateError) {
+      [strongDelegate willPresentFullScreenView];
+      NSError *presentError = GADUnityErrorWithDescription(@"Finish State Error");
```
ericleich: In the error state, is the "start" callback not called? Or could this be any error during presentation?

Wer want to make sure we only call willPresentFullScreenView once.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -42,189 +33,59 @@ + (instancetype)sharedInstance {
...
-  // should use the unityAdsReady: and unityAdsDidError: callbacks to forward Unity Ads SDK state to
-  // Google Mobile Ads SDK.
+    if ((newState == kUnityAdsPlacementStateWaiting || newState == kUnityAdsPlacementStateReady) && self.completeBlock) {
+      self.completeBlock(nil, @"UnityAds Initialization Succeeded");
+      self.completeBlock = nil;
+    }
+}
+
+- (void)unityAdsDidError:(UnityAdsError)error withMessage:(NSString *)message {
+    if (error == kUnityAdsErrorNotInitialized || error == kUnityAdsErrorInvalidArgument || error == kUnityAdsErrorInitializedFailed || error == kUnityAdsErrorInitSanityCheckFail) {
```
ericleich: Why gate on specific errors? Why wouldn't we always want to forward the error?

jindoujiao:
Hi @ericleich @joshuajplagonera,
I refactored codes to bring back Singleton class requestAd and presentAd methods to avoid loading same placementID from multiple adapters. 
I also changed the unityAdsDidError() implementation: handle two categories errors: initialization errors and show time errors. 

Please review the latest code. Thank you !

jindoujiao:
@googlebot I consent

jindoujiao:
@googlebot I consent.

AliceNoWonderland:
I consent.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -203,22 +180,22 @@ - (void)unityAdsDidClick:(NSString *)placementID {
...
-
-  NSArray *delegates;
-  @synchronized(_adapterDelegates) {
-    delegates = _adapterDelegates.objectEnumerator.allObjects;
-  }
-
-  for (id<UnityAdsExtendedDelegate, UnityAdsExtendedDelegate> delegate in delegates) {
-    [delegate unityAdsDidError:error withMessage:message];
-  }
+  if (error == kUnityAdsErrorInitSanityCheckFail || error == kUnityAdsErrorNotInitialized || error == kUnityAdsErrorInvalidArgument ||
```
ericleich: Will the UnityAds SDK invoke unityAdsDidError:withMessage: for each call to [UnityAds initialize:testMode:enablePlacementPerLoad:]? I ask because, once we get this failure once, any future ad loads will still see initialized to false and try to initialize again. I want to make sure we're waiting on a callback that will actually come, and not just time out.

If we can't expect multiple callbacks, the adapter should keep track of its own initialization state to track this.

Please confirm, this is essentially my only blocking comment.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
```diff
@@ -154,26 +135,22 @@ - (void)presentInterstitialAdForViewController:(UIViewController *)viewControlle
...
 #pragma mark - Unity Delegate Methods
 
 - (void)unityAdsPlacementStateChanged:(NSString *)placementId
                              oldState:(UnityAdsPlacementState)oldState
                              newState:(UnityAdsPlacementState)newState {
-  // This callback is not forwarded to the adapter by the GADMAdapterUnitySingleton and the adapter
-  // should use the unityAdsReady: and unityAdsDidError: callbacks to forward Unity Ads SDK state to
-  // Google Mobile Ads SDK.
+  id<GADMAdapterUnityDataProvider, UnityAdsExtendedDelegate> adapterDelegate;
+  @synchronized(_adapterDelegates) {
```
ericleich: No changes required now for this since the adapter uses synchronized elsewhere, but heads up that when we have more time at a later date we will consider refactoring this code to use dispatch per best practice #2 on https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/CONTRIBUTING.md#best-practices

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
ericleich: Not sure what I meant by this comment,  I think I made this before code changed. Consider this resolved.

adapters/Unity/UnityAdapter/GADMAdapterUnityBannerAd.m
```diff
@@ -57,7 +57,7 @@ - (void)loadBannerWithSize:(GADAdSize)adSize {
 
   if (![UnityAds isInitialized]) {
     NSString *gameID = [strongConnector.credentials[kGADMAdapterUnityGameID] copy];
-    [GADMAdapterUnitySingleton.sharedInstance initializeWithGameID:gameID];
+      [[GADMAdapterUnitySingleton sharedInstance] initializeWithGameID:gameID];
```
ericleich: Any reason to not use the dot notation? We typically prefer it as it's less brackets.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -35,6 +34,9 @@ @interface GADMAdapterUnityRewardedAd () <GADMAdapterUnityDataProvider, UnityAds
 
   /// YES if the adapter is loading.
   BOOL _isLoading;
+  
+  NSString *_uuid;
```
ericleich: Please comment fields.

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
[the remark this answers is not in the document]
AliceNoWonderland: Sounds good, we will also be planning in releasing a more dedicated show error API in the future , but it won't be in our next major release

AliceNoWonderland: Yes, any failed initialization will trigger these errors

AliceNoWonderland:
@ericleich  Please take a look about the latest changes

adapters/Unity/UnityAdapter/GADMAdapterUnitySingleton.m
ericleich: Thanks for confirming!

ericleich:
@joshuajplagonera please run some sanity tests and then we can merge. Thanks for making the requested changes @AliceNoWonderland!

PR #197, opened 2020-02-24
Vungle iOS AdMob Adapter 6.5.2.0 Release

ashinagawa:
This build is compatible with Vungle iOS SDK 6.5.2.


ashinagawa:
@googlebot I consent.

YueVungle:
@googlebot I consent.

YueVungle:
@googlebot I fixed it.

joshuajplagonera:
Hi @YueVungle 
Can you fetch and pull the latest changes from the repository?

adapters/Vungle/CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 ## Vungle iOS Mediation Adapter Changelog
 
+#### Version 6.5.2.0
+- Verified compatibility with Vungle SDK 6.5.2.
+- Added support for Smart and Adaptive Banner ads.
+
+#### Version 6.5.1.1
```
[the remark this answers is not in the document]
YueVungle: @joshuajplagonera  Got  it, I will remove  `6.5.1.1` version and combine all the notes under  `6.5.2.0`. Thanks a lot for the instruction. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -103,39 +108,70 @@ - (BOOL)isSDKInitialized {
   return [[VungleSDK sharedSDK] isInitialized];
 }
 
-- (BOOL)isAdCachedForPlacementID:(nonnull NSString *)placementID {
-  return [[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID];
+- (BOOL)isAdCachedForPlacementID:(nonnull NSString *)placementID withDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
+  GADMAdapterVungleAdType adType = [delegate adapterAdType];
+  if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner && adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
```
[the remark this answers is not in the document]
YueVungle: @joshuajplagonera Yes, MREC is not included here, because  MREC only has one fixed size.  If the delegate is MREC, the `isBannerAd` method returns YES, because MREC could be seen as a kind of Banner to some extent.  Thanks. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -365,11 +496,72 @@ - (void)vungleSDKDidInitialize {
...
+  [self initialized:YES error:nil];
 }
 
 - (void)vungleSDKFailedToInitializeWithError:(nonnull NSError *)error {
   [self initialized:false error:error];
 }
 
 @end
+
+@interface GADMAdapterVungleBannerRequest ()
```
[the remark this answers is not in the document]
YueVungle: @joshuajplagonera Got it, I will put `GADMAdapterVungleBannerRequest` class in a separate file. Thanks a lot. ^_^

YueVungle:
> Hi @YueVungle
> Can you fetch and pull the latest changes from the repository?
@joshuajplagonera Got it, I will fetch and pull the latest changes from the repository. Thanks a lot for the instruction. 🙂

adapters/Vungle/CHANGELOG.md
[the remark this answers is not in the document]
YueVungle: @joshuajplagonera  Done, thanks a lot. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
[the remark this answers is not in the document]
YueVungle: @joshuajplagonera  I have finished putting  `GADMAdapterVungleBannerRequest` class in a separate file according to your instruction. Thanks a lot. 🙂

YueVungle:
> Hi @YueVungle
> Can you fetch and pull the latest changes from the repository?
Done, thanks a lot. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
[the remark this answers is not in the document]
YueVungle: @joshuajplagonera I'm really sorry, maybe my last comment is not clear. If we use `[delegate isBannerAd]` here,  this could lead to MREC call `- (BOOL)isAdCachedForPlacementID:withSize:` method, which MREC should call `- (BOOL)isAdCachedForPlacementID:` method. Because if the delegate is MREC, the `[delegate isBannerAd]` returns YES, since MREC could be seen as a kind of Banner to some extent. Thanks a lot. 😊

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
```diff
@@ -37,4 +37,8 @@
 
 @property(nonatomic, copy) NSString *_Nullable playingPlacement;
 
+@property(nonatomic, copy) NSNumber *_Nullable orientations;
+
+@property(nonatomic, copy) NSString *_Nonnull UUID;
```
ericleich: This must be readonly so nobody can change it.

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -145,18 +167,19 @@ - (void)getInterstitial {
     [strongConnector adapter:self didFailAd:error];
     return;
   }
-  [[GADMAdapterVungleRouter sharedInstance] initWithAppId:appID delegate:self];
+  GADMAdapterVungleInterstitial *__weak weakSelf = self;
```
ericleich: This doesn't do anything. Just because this reference is weak, it's still pointing to the object. If the class it's passed to keeps this as strong, it will still be a strong reference.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -55,7 +58,7 @@ - (id)init {
     _initializingDelegates = [NSMapTable mapTableWithKeyOptions:NSPointerFunctionsStrongMemory
                                                    valueOptions:NSPointerFunctionsWeakMemory];
     _bannerDelegates = [NSMapTable mapTableWithKeyOptions:NSPointerFunctionsStrongMemory
-                                             valueOptions:NSPointerFunctionsWeakMemory];
+                                             valueOptions:NSPointerFunctionsStrongMemory];
```
ericleich: This should not be strong. Again, when this class is passed an object and keeps a strong reference to it, it doesn't matter if the caller passed in a __weak poiniter.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -103,37 +109,85 @@ - (BOOL)isSDKInitialized {
   return [[VungleSDK sharedSDK] isInitialized];
 }
 
-- (BOOL)isAdCachedForPlacementID:(nonnull NSString *)placementID {
-  return [[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID];
+- (BOOL)isAdCachedForPlacementID:(nonnull NSString *)placementID withDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
+  GADMAdapterVungleAdType adType = [delegate adapterAdType];
+  if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner && adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
+    return [[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID];
```
ericleich: What would Vungle return if calling this method with a banner placement here without a size? Return true if any size is cached for that banner?

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -25,7 +28,7 @@ @implementation GADMAdapterVungleRouter {
   NSMapTable<NSString *, id<GADMAdapterVungleDelegate>> *_initializingDelegates;
 
   /// Map table to hold the banner ad delegates with placement ID as a key.
-  NSMapTable<NSString *, id<GADMAdapterVungleDelegate>> *_bannerDelegates;
+  NSMapTable<GADMAdapterVungleBannerRequest *, NSMutableOrderedSet<id<GADMAdapterVungleDelegate>> *> *_bannerDelegates;
```
ericleich: Why is the value in this MapTable now a set? Doesn't Vungle still only support a single banner per placement ID? This makes all of the code dealing with this set more complicated than it needs to be.

Or does Vungle only support 1 banner ad at a time period (even 2 banners with unique placement IDs not supported)? Based on implementation of canRequestBannerAdForPlacementID: and the fact that the entire router class only holds 1 _bannerPlacementDI, it seems like only 1 banner is supported. If that's the case, this code can be greatly simplified by removing this _bannerDelegates altogether, replacing _bannerPlacementID to be a GADMAdapterVungleBannerRequest object, and immediately denying any requests coming in if _bannerPlacementID is not nil and doesn't match the UUID of the next request (possible refresh) coming in.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -261,7 +376,12 @@ - (nullable UIView *)renderBannerAdInView:(nonnull UIView *)bannerView
                                    extras:(nullable VungleAdNetworkExtras *)extras
                            forPlacementID:(nonnull NSString *)placementID {
   NSError *bannerError = nil;
-  NSMutableDictionary *options = [[NSMutableDictionary alloc] init];  ///
+  NSMutableDictionary *options = [[NSMutableDictionary alloc] init];
+  if (extras.muted) {
```
ericleich: This line should be:

if (extras != nil)

The way it's written now, you're setting muted to YES every time.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -288,12 +408,23 @@ - (nullable UIView *)renderBannerAdInView:(nonnull UIView *)bannerView
   return nil;
 }
 
-- (void)completeBannerAdViewForPlacementID:(nullable NSString *)placementID {
-  if (_isBannerPresenting) {
-    NSLog(@"Vungle: Triggering an ad completion call for %@", placementID);
-    _isBannerPresenting = NO;
+- (void)completeBannerAdViewForPlacementID:(nonnull id<GADMAdapterVungleDelegate>)delegate; {
```
ericleich: Should not be a ; before that opening bracket. I suspect this might break compilation.

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
YueVungle: @ericleich I thought pubs could set the value of `UUID` to differentiate Banners withe same unitId. Got it, I will update `UUID` as a `readonly` property. Thanks a lot for the instruction. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
YueVungle: @ericleich Got it, I will remove this redundant implementation. Thanks a lot for your guidance.  ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Yes, Vungle only supports 1 Banner per placement ID at a time even 2 Banners with unique placement ID does not support. After one Banner ad is closed(the _bannerPlacementID will be set to nil), the SDK could load and display another placement ID Banner.

But When the old Banner request object is displaying a Banner ad, the refresh Banner request object is loading a new Banner ad and also needs to be added to the _bannerDelegates, when the refresh Banner request object constructs successfully, before adding to the screen, the old Banner request object will be closed and removed from _bannerDelegates. During this period, there are two Banner request objects in _bannerDelegates. Our 6.5.2.0 Adapter updates the logic to track Banner request object lifecycle, after calling the delegate method `vungleDidCloseAdWithViewInfo:placementID:`, then the adapter could remove the old Banner request object  from _bannerDelegates.

The previous implementation is that one placementID only corresponds to one delegate, which could lead the refresh request delegate cannot be added to _bannerDelegates, after the old delegate is removed from _bannerDelegates, the _bannerDelegates is empty, so the the refresh request delegate cannot receive any callbacks. Thanks a lot. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi ericleich, we used the weak reference before, but in the testing of multiple Banners with same placementID, `NSPointerFunctionsWeakMemory` always leads the _bannerDelegates becoming empty, because the value object of _bannerDelegates is released early. And our 6.5.2.0 Adapter updates the logic to track Banner delegate lifecycle, after received the callback method `vungleDidCloseAdWithViewInfo:placementID:`, then the adapter will remove the old Banner request delegate from _bannerDelegates, so use strong reference here is fine. And I will remove that redundant __weak assignment operation. Thanks a lot for the instructions. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Yes, it will return true if any size is cached for that banner. Thanks a lot O(∩_∩)O

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Got it, I will fix this issue, thanks a lot for the guidance. 👍

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Thanks so much for pointing out this issue. I will remove ; before the opening bracket. And have a nice day. ^_^

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
ericleich: @YueVungle my original proposal was something that publishers set, but your idea of the adapter automatically generating a UUID is a better way of guaranteeing that every instance of extras has a unique identifier. So publishers no longer need to effectively figure out for themselves how to provide a unique ID. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: This is kind of related to the comment I'll send below. I think moving away from a map would will resolve this.

The reason it's getting cleared is because we moved to the value being an ordered set of delegates, not just a single delegate. Back when it was a single delegate, the Google SDK would hold onto the adapter instance if it was still valid, and it would allow the adapter (aka the delegate) to nil out if Google no longer considered this instance valid (such as a request timeout). It's good practice to preserve this behavior via weak references -- ideally Google SDK keeps the only strong reference to the adapter, so that if it gets deallocated (e.g. timeout) and any code path for some reason doesn't call the router to remove the banner delegate, it gets automatically cleared.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: Understand the reasoning for the list of adapters for a single placement. But given that Vungle can currently only load 1 ad per placement, I think we can just get rid of the HashMap though and support just a single array of adapters. And per my comment above, leverage [NSPointerArray](https://developer.apple.com/documentation/foundation/nspointerarray?language=objc) with an option for NSPointerFunctionsWeakMemory to its delegate objects.

This should greatly simplify a lot of the methods. And if a new request comes and the array has a non-nil object with a different placement ID or UUID, we fail the request.

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
YueVungle: @ericleich Got it, thanks a lot for the detailed explanation. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks a lot for your guidance. I have a concern, if we move away from a map and use NSPointerArray with an option for NSPointerFunctionsWeakMemory to manage Banner delegates. During my debugging, I found that AdMob SDK always removes old Banner delegate and set it to nil before Vungle Adapter receives the callback method
`- (void)vungleWillCloseAdWithViewInfo:placementID:` during refreshing process,  this could lead the old Banner delegate cannot receive `- (void)willCloseAd:didDownload:` delegate method. This is why we update the logic to track Banner delegate lifecycle, after Vungle Adapter receives the callback method `- (void)vungleDidCloseAdWithViewInfo:placementID:', the adapter will remove the old Banner delegate from _bannerDelegates. Thanks a lot.  :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks a lot for your guidance. I have a concern, if we move away from a map and use NSPointerArray with an option for NSPointerFunctionsWeakMemory to manage Banner delegates. During my debugging, I found that AdMob SDK always removes old Banner delegate and set it to nil before Vungle Adapter receives the callback method
`- (void)vungleWillCloseAdWithViewInfo:placementID:` during refreshing process, this could lead the old Banner delegate cannot receive `- (void)willCloseAd:didDownload:` delegate method. This is why we update the logic to track Banner delegate lifecycle, after Vungle Adapter receives the callback method `- (void)vungleDidCloseAdWithViewInfo:placementID:` , the adapter will remove the old Banner delegate from _bannerDelegates. Thanks a lot. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks a lot for your instruction. But in the coming version of Vungle SDK, multiple Banners will be supported. If we remove NSMapTable and use NSPointerArray in current version, we need to add NSMapTable back in the coming version to manage displaying multiple Banners at the same time.  😊

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
YueVungle: @ericleich  Done, thanks. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
YueVungle: @ericleich  Done, thanks for your instruction. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich  Done, thanks for your instruction. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich  Done, thanks a lot. 😊

YueVungle:
@ericleich  Hi Eric, sorry to make you confusing. Because Vungle released 6.5.3 SDK, the Vungle Adapter also updated to 6.5.3.0(6.5.3.0 Adapter includes some bug fixing and removing i386 architecture). Since Vungle 6.5.2.0 PR review is already in progress, I didn't close this PR and added the 6.5.3.0 changes to this PR directly so that we don’t have to review the long PR again. If you think I should close this PR and submit a new PR for 6.5.3.0 Adapter, please tell me and I will follow the guidance. Thanks so much. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: @YueVungle If Google SDK is dropping the instance, it means we also dropped the ad that the adapter is responding to. So the callback likely isn't doing anything anyways (or if it's making it back, it's only because the adapter didn't let things get cleaned up). I took a look at what `(void)willCloseAd:didDownload:` does and it only seemed to be calling a callback, nothing else that would break functionality.

That said, this discussion may be a moot point if we stick with NSMapTable.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: Indeed, better to plan for the future then. Let's keep NSMapTable.

I'd still actually like to reopen the discussion of needing the array though. In your comment above, you mentioned it was due to losing a callback, but I don't think that callback was doing anything important and it's a sign that the publisher is no longer using that banner.

That said, would `NSMapTable<GADMAdapterVungleBannerRequest *, id<GADMAdapterVungleDelegate>>` still work?

ericleich:
@YueVungle Thank you for keeping the same PR. This is definitely preferred!

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks a lot for the instructions, I see and I think we should remove the logic to track Banner delegate lifecycle, let Google SDK manage the Banner delegate lifecycle. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, I agree with you.  If Google SDK is dropping the instance, it means it also dropped the ad that the adapter is responding to. So I think we don't need to track the `- (void)vungleDidCloseAdWithViewInfo:placementID: ` callback if the Banner delegate is set to nil by Google SDK . We should let Google SDK manage the Banner delegate lifecycle.  I will change the `_bannerDelegates` back to use `NSMapTable<GADMAdapterVungleBannerRequest *, id<GADMAdapterVungleDelegate>>`, thanks so much for all the help and instructions. Have a nice day. 🙂

YueVungle:
@ericleich Hi Eric, we have changed the `_bannerDelegates` back to use `NSMapTable<GADMAdapterVungleBannerRequest *, id<GADMAdapterVungleDelegate>>` and for a refresh banner delegate, when the Banner view is constructed successfully, it will replace the old banner delegate in `_bannerDelegates`. Please review the changes, thanks so much and have a nice day. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, we changed the `_bannerDelegates` back to use `NSMapTable<GADMAdapterVungleBannerRequest *, id<GADMAdapterVungleDelegate>>` and used `NSPointerFunctionsWeakMemory`, thanks so much for your instruction. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, we changed the `_bannerDelegates` back to use `NSMapTable<GADMAdapterVungleBannerRequest *, id<GADMAdapterVungleDelegate>>`, thanks a lot for your guidance. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -126,67 +126,87 @@ - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
+        _bannerRequest = [delegate.bannerRequest copy];
+      } else {
+        id<GADMAdapterVungleDelegate> bannerDelegate = [_bannerDelegates objectForKey:delegate.bannerRequest];
+
+        /* The isRequestingBannerAdForRefresh flag is used for an edge case, when the old Banner
+         * delegate is removed from _bannerDelegates and there is a refresh Banner delegate doesn't
+         * construct Banner view successfully and add to _bannerDelegates yet. Adapter cannot set
+         * _bannerRequest to nil to prevent requesting another Banner ad with different bannerRequest.
+         */
+        bannerDelegate.isRequestingBannerAdForRefresh = YES;
```
ericleich, line 167: It took me a while to understand why you had these two booleans but I get it now. However, I think you could accomplish the same thing without those 2 boolean flags:

In this method, you could overwrite the NSMapTable with the new delegate after confirming it's the same banner request.

Then, in the removeDelegate method, you could write:

    if (delegate && (delegate == [_bannerDelegates objectForKey:delegate.bannerRequest])) {
      return;
    }
    GADMAdapterVungleMapTableRemoveObjectForKey(_bannerDelegates, delegate.bannerRequest);

Since the refreshed ad will already have taken over in the NSMapTable, when the old delegate tries to remove itself, it wouldn't pass the `delegate == [_bannerDelegates objectForKey:delegate.bannerRequest])` check. And it wouldn't get removed. Additionally, _bannerRequest stays intact the entire time.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -126,67 +126,87 @@ - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
+          // but different uniquePubRequestID in _bannerDelegates.
+          if ([vungleBannerRequest.placementID isEqualToString:delegate.bannerRequest.placementID]) {
             return NO;
-          } else {
-            NSOrderedSet *set = [_bannerDelegates objectForKey:bannerRequest];
-            for(id<GADMAdapterVungleDelegate> delegate in set) {
-              if (delegate.bannerState == BannerRouterDelegateStateWillPlay || delegate.bannerState == BannerRouterDelegateStatePlaying) {
-                return NO;
-              }
+          } else if (vungleBannerRequest.placementID.length > 0 && ![vungleBannerRequest.placementID isEqualToString:delegate.bannerRequest.placementID]) {
```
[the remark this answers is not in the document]
YueVungle: @ericleich  Hi Eric, thanks a lot for pointing out this issue, we should check `placementID.length > 0` earlier in VungleInterstitial class, if you don't have time to refactor this issue, we will fix it in the next version.  ：）

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -126,67 +126,87 @@ - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
+        bannerDelegate.isRequestingBannerAdForRefresh = YES;
+        delegate.isRefreshedForBannerAd = YES;
       }
     }
   }
   return YES;
 }
 
+- (void)replaceOldBannerDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
+  if (!delegate) {
```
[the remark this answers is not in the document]
YueVungle: @ericleich Hi Eric, yes, this should never be nil, I will remove this check. Thanks a lot for the guidance. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -126,67 +126,87 @@ - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
       }
 
-      if (delegate && ![_bannerDelegates objectForKey:delegate.bannerRequest]) {
+      if (![_bannerDelegates objectForKey:delegate.bannerRequest]) {
         NSEnumerator *enumerator = _bannerDelegates.keyEnumerator;
-        GADMAdapterVungleBannerRequest *bannerRequest = nil;
-        while (bannerRequest = [enumerator nextObject]) {
-          if ([bannerRequest.placementID isEqualToString:delegate.bannerRequest.placementID]) {
+        GADMAdapterVungleBannerRequest *vungleBannerRequest = nil;
+        while (vungleBannerRequest = [enumerator nextObject]) {
```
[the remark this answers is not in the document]
YueVungle: @ericleich Hi Eric, thanks a lot for the suggestion, we will use  placement ID as the key and use VungleBannerRequest as a property of the delegate in next version, this is more effective way. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, if we overwrite the NSMapTable with the new delegate after confirming it's the same banner request before the new refresh banner view is constructed successfully, this will cause problems in our next version which supports new event handlers, such as click event. If users click the old banner view, then adapter cannot send the click event to the old delegate which is replaced by the new refresh banner delegate. Thanks a lot. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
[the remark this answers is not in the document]
YueVungle: @ericleich Done, thanks a lot for the instruction. 🙂

YueVungle:
> @YueVungle this is about ready but I have a few more code health suggestions. Please confirm. My goal is to accept after this last round of feedback, with any remaining code health comments to be addressed later. The main one i'd like to address now is if we really need those 2 new booleans.
> 
> @joshuajplagonera please also run tests against the latest code.

@ericleich Hi Eric, thanks so much for pointing out these code health suggestions, we will fix them in the next version. And for 2 new booleans, I have a concern about overwriting the NSMapTable with the new delegate after confirming it's the same banner request in our next version which supports new event handlers, such as click event. Please take a look at the related code comment. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -250,7 +267,7 @@ - (BOOL)hasDelegateForPlacementID:(nonnull NSString *)placementID
 - (BOOL)canRequestBannerAdForPlacementID:(nonnull GADMAdapterVungleBannerRequest *)bannerRequest {
   @synchronized(_bannerDelegates) {
     if (_bannerDelegates.count > 0) {
-      return _bannerPlacementID == nil || [_bannerDelegates objectForKey:bannerRequest];
+      return _bannerRequest == nil || [_bannerRequest isEqualToBannerRequest:bannerRequest];
```
ericleich, line 266: _bannerRequest should never be nil, right? Because what's passed in is nonnull. A nil _bannerRequest would signify a bigger issue and shouldn't request ads.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -126,67 +126,87 @@ - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
 
+- (void)replaceOldBannerDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
+  if (!delegate) {
+    return;
+  }
+
+  if ([delegate respondsToSelector:@selector(isBannerAd)] && [delegate isBannerAd]) {
+    @synchronized(_bannerDelegates) {
+      // We only support displaying one Vungle Banner Ad at the same time currently
+      if (_bannerRequest != nil && ![_bannerRequest isEqualToBannerRequest:delegate.bannerRequest]) {
```
ericleich, line 179: At this point where this method is called, we've already allowed a new ad to load. If we quit out here and don't let the delegate get overridden, then we'll be sending events to the old delegate instance even though this is for the ad refresh. I think it's too late to make this check here.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -126,67 +126,87 @@ - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
+         */
+        bannerDelegate.isRequestingBannerAdForRefresh = YES;
+        delegate.isRefreshedForBannerAd = YES;
       }
     }
   }
   return YES;
 }
 
+- (void)replaceOldBannerDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
```
ericleich, line 175: nit: Suggest renaming to:

replaceOldBannerDelegateWithDelegate:

since what's getting passed in is not the old banner delegate but actually the new one :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: Ok, let's keep this then. In theory, it shouldn't matter because this should be an ad refresh, so the old instance is going away anyways. However, if Vungle ad fails to load and ultimately the request is a no fill, the old ad will still be there. So better to not remove the delegate.

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -56,34 +60,55 @@ - (void)dealloc {
...
     return;
   }
 
+  VungleAdNetworkExtras *networkExtras = [strongConnector networkExtras];
   self.desiredPlacement = [GADMAdapterVungleUtils findPlacement:[strongConnector credentials]
-                                                  networkExtras:[strongConnector networkExtras]];
-
+                                                  networkExtras:networkExtras];
+  self.bannerRequest = [[GADMAdapterVungleBannerRequest alloc] initWithPlacementID:self.desiredPlacement ?: @""
+                                                                uniquePubRequestID:networkExtras.UUID ?: @""];
```
ericleich, line 100: @YueVungle we caught an issue here during testing. If 2 different banners are loaded with the same placement ID (not a refresh) and don't specify a uniquePubRequestID, both try to load but only 1 works.

Per email discussions, we agreed that by default, if uniqueRequestID isn't specified, we couldn't detect if it was a refresh and we would fail the ad.

Ultimately, I suggest not defaulting to @"" here (just take the result of networkExtras.UUID which may be nil), and changing this parameter to be nullable inside GADMAdapterVungleBannerRequest. Then when comparing GADMAdapterVungleBannerRequest objects, comparing [nil isEqualToString:nil] will fail if 2 banner requests don't specify this extra. Currently it's defaulting to a pass because [@"" isEqualToString:@""] returns YES.

This is the one remaining issue we'd like fixed prior to approvinig.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich  Hi Eric, when a Banner finished displaying ad such as users close the Banner ad view or exit the VC displays a Banner ad, _bannerRequest should be nil. Then adapter can request and display another Banner ad that has a different bannerRequest. Thanks a lot. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks for the good suggestion,  I agree with you. This is just a safety check, just don't want to add a delegate with different bannerRequest. For a refresh banner delegate, that rarely happens, because the _bannerRequest should be always equal to the bannerRequest of a refresh banner delegate.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Got it, I will change the name to `replaceOldBannerDelegateWithDelegate:`. Thanks a lot for the guidance. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
YueVungle: @YueVungle Hi Eric, I know Vungle Android side already implemented above suggestion. We will implement above suggestion on iOS side too . But I have a concern about above implementation, if there is only one Banner ad in a page and doesn't register networkExtras(so uniquePubRequestID is nil), this could lead the refresh Banner request fail to load. Thanks O(∩_∩)O

ericleich: @YueVungle yes that is true, but no uniquePubRequestID also means we can't detect if it was 2 different requests. So we'd be making an assumption that it's a refresh without knowing for sure. And per our email conversation, it's a much bigger publisher experience issue to accidentally assume a refresh when it isn't, because the first ad would then be empty. So it's important to be conservative here and drop that request.

On Android, we had the adapter log a clear message when this happens, to help notify publishers of this case:

`Ad already loaded for placement ID: " + placementId + ", and cannot determine if this is a refresh. Set Vungle extras when making an ad request to support refresh on Vungle banner ads.`

YueVungle: @ericleich Hi Eric, got it, thanks a lot for the detailed explanation and guidance, we will implement above suggestion. Have a great day! ^_^

ericleich: @YueVungle I think we're still waiting on getting the updated code after you finish testing. Please let me know once it's ready!

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Done, thanks a lot. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
YueVungle: @ericleich Hi Eric, I'm really really sorry for the late response,  because before committing the changes of this suggestion about refresh checking, the changes must pass internal PR review and QA/IE test. And QA/IE colleagues have other tests scheduled in advance, so the QA test just passed. I have updated the codes, now if uniqueRequestID isn't specified(it is set to nil), we couldn't detect if it was a refresh and we would fail the ad. Please review the changes and thanks so much for all the help and guidance. Have a great day! ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -229,13 +230,13 @@ - (void)removeDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
...
     @synchronized(_bannerDelegates) {
       if (delegate && (delegate == [_bannerDelegates objectForKey:delegate.bannerRequest])) {
-        if ([delegate.bannerRequest isEqualToBannerRequest:_bannerRequest]) {
-          if (!_isBannerPresenting && !delegate.isRequestingBannerAdForRefresh) {
-            _bannerRequest = nil;
-          }
-        }
         GADMAdapterVungleMapTableRemoveObjectForKey(_bannerDelegates, delegate.bannerRequest);
       }
+      if ([delegate.bannerRequest isEqualToBannerRequest:_bannerRequest] || ([delegate.bannerRequest.placementID isEqualToString:_bannerRequest.placementID] && !delegate.bannerRequest.uniquePubRequestID && !_bannerRequest.uniquePubRequestID)) {
```
ericleich, line 235: @YueVungle I'm not sure why this needs to change, since in other places where we check if we can load an ad again, we still only rely on [delegate.bannerRequest isEqualToBannerRequest:_bannerRequest]. Having said that, I'm not sure what the side effects are so I won't press on this. This looks good to me and we'll merge pending final testing from @joshuajplagonera

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
ericleich: @YueVungle no worries, just wanted to make sure you weren't just waiting for me to review.

The changes are ready to merge pending final testing from @joshuajplagonera . Thanks!

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich  Hi Eric, I making this change is for the scenario that when the refresh Banner request is failed to load ad and not added to _bannerDelegates, then users close this Banner view. The adapter should need to set _bannerRequest to nil to let the users can load another Banner ad with different GADMAdapterVungleBannerRequest. Thanks a lot. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
YueVungle: @ericleich Hi Eric, got it, big thanks to you for all the quick reply, help and guidance. I really appreciate it. 😊

PR #199, opened 2020-03-04
Use the Facebook native ad's aspect ratio

nemdub:
Because _mediaView.aspectRatio is available only once the media view is rendered in a view.

This created a problem where the aspectRatio value would change from 0 to an actual value depending on whether the mediaView was rendered or not.


nemdub:
@googlebot I signed it!

joshuajplagonera:
Hi @nemdub 
Please pull in the latest changes and resolve any merge conflicts.

nemdub:
done :)

PR #202, opened 2020-03-18
Chartboost 8.1.0 support

daniel-barros:
SDK 8.1.0 introduces new ad types and SDK initialization method, which require a huge rewrite of the adapters to support.
Main changes:
- Improved separation of concerns with GADChartboostSingleton, GADCHBInterstitial, GADCHBRewarded and GADCHBBanner.
- Many simplifications due to Chartboost dropping its singleton pattern for dealing with ads, and having a start method with a completion block.


joshuajplagonera:
Hi @daniel-barros 
Can you take a look at https://github.com/googleads/googleads-mobile-ios-mediation/pull/224 ?

PR #207, opened 2020-03-30
Updates to Verizon SE SDK v1.5.0

lcsabaignat:


adapters/VerizonMedia/VerizonMediaAdapter/Native/GADMAdapterVerizonNativeAd.m
```diff
@@ -54,7 +54,7 @@ - (void)loadNativeAdWithAdTypes:(nonnull NSArray<GADAdLoaderAdType> *)adTypes
     return;
   }
   _nativeAdFactory = [[VASNativeAdFactory alloc] initWithPlacementId:_placementID
-                                                             adTypes:@[ @"inline" ]
+                                                             adTypes:@[ @"100", @"simpleImage" ]
```
joshuajplagonera: Can you explain what this changes?

thomas-poland: @joshuajplagonera Sure - Its an update to our native types.  Previously, 100 mapped to the simpleImage type.  We're just renaming them.  

I'm working with @lcsabaignat on this; he will address the change log edit; I'm investigating the Thread Checker issue.

thomas-poland: @joshuajplagonera I am able to reproduce the thread check error, which I'm continuing to investigate. However - this should not prevent rewarded video fills, that is working fine in our Test Harness - please can you provide more information  - is the VAS SDK logging an error for that ad request?   thanks

joshuajplagonera: Hi Thomas,
We are receiving the following error from the `interstitialAdFactory:didFailWithError:` callback:

`Error Domain=com.verizon.ads.vastController.ErrorDomain Code=2 "Load timed out." UserInfo={VASErrorInfoWho=VASVASTController, NSLocalizedDescription=Load timed out.}`

joshuajplagonera: Hi Thomas,
Just an update, I was able to get Rewarded ads from Verizon after many retries, so it is safe to say that the Thread Check error is not responsible for the fill issue.

thomas-poland: @joshuajplagonera also, you could increase the timeout by adding this line at the beginning of the  `- (void)loadRewardedAdForAdConfiguration:(nonnull GADMediationRewardedAdConfiguration *)adConfig...` method:


``` [[[VASAds sharedInstance] configuration] setInteger:30000 forDomain:kDomainVASInterstitialPlacement key:kVASInterstitialAdAdapterTimeoutKey security:nil];```

I'm suggesting this rather than updating the PR so that you can try it today instead of waiting for another time zone cycle (Lorand is in Eu); we can add it if you confirm that is helpful.

joshuajplagonera: Hi Thomas,
Thanks for the additional information. However I'm not able to increase the timeout as you advised since the `domain` and `key` used is unknown. Would you know which file we need to import in order to use those keys?

That said, I was able to get fill for Rewarded Ads sometimes, so it does not always fail. However, here is the information about the placement.
Site ID: `8a96911a016a6a6a46d26b14520700fb`
Placement: `rewarded`

thomas-poland: @joshuajplagonera Based on sometimes filling, that is almost certainly the timeout - my bad, cut and paste error in my edit above, here are the missing constants:

```static NSString * const kDomainVASInterstitialPlacement         = @"com.verizon.ads.interstitialplacement";```

```static NSString * const kVASInterstitialAdAdapterTimeoutKey     = @"interstitialAdAdapterTimeout";```

Once we agree on a reasonable value, we'll update the Adapter appropriately and @lcsabaignat will update the PR.

joshuajplagonera: Hi Thomas,
I synced with the team internally and it was agreed that the adapter should stick to what the default timeout is for the Verizon SDK, and the adapter shouldn't override it.

That said, the PR looks good to me, and I will continue with the merge after certifying the changes. 

Thanks!

PR #218, opened 2020-04-20
Verified MyTarget iOS adapter compatibility with MyTarget iOS SDK version 5.5.1

aserrr:


aserrr:
@googlebot I fixed it.

PR #219, opened 2020-04-20
Verified MyTarget iOS adapter compatibility with MyTarget iOS SDK version 5.5.1

aserrr:


adapters/MyTarget/MyTargetAdapter/GADMAdapterMyTargetMediatedUnifiedNativeAd.m
```diff
@@ -31,13 +31,14 @@ @implementation GADMAdapterMyTargetMediatedUnifiedNativeAd {
...
     mediatedUnifiedNativeAdWithNativePromoBanner:(nonnull MTRGNativePromoBanner *)promoBanner
                                         nativeAd:(nonnull MTRGNativeAd *)nativeAd
-                                  autoLoadImages:(BOOL)autoLoadImages
                                      mediaAdView:(nonnull MTRGMediaAdView *)mediaAdView {
   if (!promoBanner.title || !promoBanner.descriptionText || !promoBanner.image ||
       !promoBanner.ctaText) {
     return nil;
   }
 
+  BOOL autoLoadImages = (nativeAd.cachePolicy == MTRGCachePolicyImages) || (nativeAd.cachePolicy == MTRGCachePolicyAll);
```
joshuajplagonera, line 40: Can we create an extras for handling autoload Images instead of using the cache policy?

aserrr: There is no need to create an extras I think. We're fill cachePolicy (depending on GADNativeAdImageAdLoaderOptions) for ad object before loading, and then check if images were actually loaded if they should.

joshuajplagonera: I believe that by making this an extra, we are making sure that autoloading images is set by the publisher and/or the adapter and not the MyTarget SDK, which may change behavior.

aserrr: But myTarget SDK doesn't set it on it's own. It uses GADNativeAdImageAdLoaderOptions which is set by publisher. Please refer to rows 123-135 in GADMAdapterMyTargetNative.m file within current PR.
That's why making an extra is useless.

aserrr: Ok. It's reasonable and acceptable.

PR #226, opened 2020-05-04
Updated for FAN 5.9.0 

yroxe:


adapters/Facebook/FacebookAdapter/GADMAdapterFacebookConstants.h
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 /// Facebook mediation network adapter version.
-static NSString *const kGADMAdapterFacebookVersion = @"5.8.0.2";
+static NSString *const kGADMAdapterFacebookVersion = @"5.9.0.1";
```
chlandacastro, line 16: version should be 5.9.0.0 instead of 5.9.0.1

yroxe: Sure, let me send an update.

PR #231, opened 2020-05-11
Fix potential adapter timeout code paths in the Vungle Adapter

ericleich:
Fix potential adapter timeout code paths in the Vungle Adapter

### DIFF
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
@@ -13,10 +13,10 @@
 // limitations under the License.
 
 #import "GADMAdapterVungleRouter.h"
+#import "GADMAdapterVungleBannerRequest.h"
 #import "GADMAdapterVungleConstants.h"
 #import "GADMAdapterVungleUtils.h"
 #import "VungleRouterConsent.h"
-#import "GADMAdapterVungleBannerRequest.h"
 
 const CGSize kVNGBannerShortSize = {300, 50};
 
@@ -95,7 +95,7 @@ const CGSize kVNGBannerShortSize = {300, 50};
   _isInitializing = YES;
   _isBannerPresenting = NO;
 
-  //Disable refresh functionality for all banners
+  // Disable refresh functionality for all banners
   [[VungleSDK sharedSDK] disableBannerRefresh];
 
   NSError *err = nil;
@@ -109,13 +109,16 @@ const CGSize kVNGBannerShortSize = {300, 50};
   return [[VungleSDK sharedSDK] isInitialized];
 }
 
-- (BOOL)isAdCachedForPlacementID:(nonnull NSString *)placementID withDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
+- (BOOL)isAdCachedForPlacementID:(nonnull NSString *)placementID
+                    withDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
   GADMAdapterVungleAdType adType = [delegate adapterAdType];
-  if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner && adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
+  if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner &&
+      adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
     return [[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID];
   }
 
-  return [[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID withSize:[self getVungleBannerAdSizeType:adType]];
+  return [[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID
+                                                withSize:[self getVungleBannerAdSizeType:adType]];
 }
 
 - (BOOL)addDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
@@ -133,7 +136,8 @@ const CGSize kVNGBannerShortSize = {300, 50};
   } else if ([delegate respondsToSelector:@selector(isBannerAd)] && [delegate isBannerAd]) {
     @synchronized(_bannerDelegates) {
       // We only support displaying one Vungle Banner Ad at the same time currently
-      if (_bannerRequest != nil && ![_bannerRequest isEqualToBannerRequest:delegate.bannerRequest]) {
+      if (_bannerRequest != nil &&
+          ![_bannerRequest isEqualToBannerRequest:delegate.bannerRequest]) {
         [self checkSamePlacementIDwithNilUniquePubRequestID:delegate.bannerRequest];
         return NO;
       }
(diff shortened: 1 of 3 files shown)


adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -284,38 +301,33 @@ - (nullable NSError *)loadAd:(nonnull NSString *)placement
...
-    }
+    return nil;
   }
 
-  return nil;
+  NSError *loadError = nil;
+  GADMAdapterVungleAdType adType = [delegate adapterAdType];
+  if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner &&
+      adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
+    [sdk loadPlacementWithID:placement error:&loadError];
```
ericleich, line 324: Vungle SDK has both a BOOL return value and an error out. Will Vungle ever return NO with a nil error, or YES with a non-nil error?

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -519,11 +544,14 @@ - (void)vungleAdPlayabilityUpdate:(BOOL)isAdPlayable
 
   if (isAdPlayable) {
     [delegate adAvailable];
-  } else if (!isAdPlayable && ![[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID]) {
+  } else {
+    // Ad not playable. Return an error.
```
ericleich, line 548: No longer checking for cached placement.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -284,38 +301,33 @@ - (nullable NSError *)loadAd:(nonnull NSString *)placement
...
+  GADMAdapterVungleAdType adType = [delegate adapterAdType];
+  if (adType != GADMAdapterVungleAdTypeBanner && adType != GADMAdapterVungleAdTypeShortBanner &&
+      adType != GADMAdapterVungleAdTypeLeaderboardBanner) {
+    [sdk loadPlacementWithID:placement error:&loadError];
+  } else {
+    [sdk loadPlacementWithID:placement
+                    withSize:[self getVungleBannerAdSizeType:adType]
+                       error:&loadError];
+  }
+  return loadError;
```
YueVungle, line 330: @ashinagawa Hi Aki, as we discussed in the email, if a placement has banner ad (320x50) cached, but SDK gets `loadPlacementWithID` call with different banner size, Vungle SDK returns "NO" and "non-nil error" for `loadPlacementWithID` call , BUT,  SDK keeps requesting a new ad for this placement, then sending "playabilityUpdate" callback. So we should keep original implementations, `if (loadError.code == VungleSDKResetPlacementForDifferentAdSize)) `, we should not return this loadError, right?  Thanks a lot. ^_^

ericleich: @YueVungle I just added a check for this above, to now revert to the old behavior. I wrote it in such a way that we can easily strip it out once Vungle SDK no longer returns this error.

Please take another look.

Sorry in advance for this still looking like 1 commit. We use [copybara](https://github.com/google/copybara) internally for this, and the tool replaces the commit when I send out for review.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, I think if `- (BOOL)loadPlacementWithID:error:` method returns NO, it will return a non-nil error, so your implementation is correct. Thanks a lot ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks for the update, it looks great. 👍O(∩_∩)O

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -519,15 +550,19 @@ - (void)vungleAdPlayabilityUpdate:(BOOL)isAdPlayable
...
-      NSLog(@"Vungle Ad Playability returned an error: %@", error.localizedDescription);
-    } else {
-      error = GADMAdapterVungleErrorWithCodeAndDescription(kGADErrorMediationAdapterError, [NSString stringWithFormat:@"Ad is not available for placementID: %@.", placementID]);
-    }
-    [delegate adNotAvailable:error];
-    [self removeDelegate:delegate];
+    return;
   }
+
+  // Ad not playable. Return an error.
```
YueVungle, line 556: @ericleich Hi Eric, I think we need to check `[[VungleSDK sharedSDK] isAdCachedForPlacementID:placementID]` here like previous implementation, because when we call `- (BOOL)playAd:options:placementID:error:` method of VungleSDK to play a fullscreen ad , before playing the ad, VungleSDK will call this `- (void)vungleAdPlayabilityUpdate:placementID:error:` delegate method and pass `NO` for the `isAdPlayable` parameter.  In this case, we should not call `[delegate adNotAvailable:error];` and `[self removeDelegate:delegate];` methods, which will cause many issues. Thanks a lot and have a great day！^_^

ericleich: @YueVungle Thanks for the feedback! I see now there is a disconnect between the router getting playabilityUpdate, and trying to tell adapters whether ad is available or not. But VungleRouter doesn't know about presentation state.

I plan to solve this by still forwarding adNotAvailable to adapters, but having the adapters check if the ad is presenting or not, and ignore the error if it's already presenting. Note: We already do the same for adAvailable in GADMAdapterVungleInterstitial and GADMAdapterVungleRewardedAd.

I'll make updates and share with you soon.

ericleich: Also FYI, GMA SDK would ignore a load error anyways if the adapter is already presenting. But it's better for that to be explicit in the adapter.

ericleich: @YueVungle code updated, please take another look.

I redefined the "presenting" status in GADMAdapterVungleInterstitial and GADMAdapterVungleRewardedAd to just be an "ad loaded" status, because we are using it to avoid sending ad load success/fail callbacks to adapters that have already loaded. This also addresses banner cases as well.

YueVungle: @ericleich Hi Eric, thanks for the update, but as I mentioned above, when we call `- (BOOL)playAd:options:placementID:error: method` of VungleSDK to play a fullscreen ad , before playing the ad, VungleSDK will call this `- (void)vungleAdPlayabilityUpdate:placementID:error:` delegate method and pass `NO` for the `isAdPlayable` parameter.  In this case, we cannot call `[self removeDelegate:delegate];`，this will prevent the delegate from receiving many important callbacks.  Have a nice Friday! O(∩_∩)O

ericleich: @YueVungle Thanks for the feedback. I added back in a check to see if the ad is cached. Please take another look.

Note: I am confused as to why the isAdCachedForPlacementID:placementID check returns YES, when isAdPlayable returns NO. The docs for that method say `Returns YES  when there is certainty that an ad will be able to play for a given placementID.`, but I'm getting a callback saying isAdPlayable=NO...

However, that does seem to be the behavior, and it would be complex to keep track of even more state in the adapter, so I stuck with that implementation.

YueVungle: @ericleich Hi Eric, thanks so much for adding the check for ad is cached back. I'm really sorry to make you confused, we will optimize our SDK in future release to make a better designed API. I saw your comments for checking ad is cached,  which is very detailed and clear. 
Now the PR looks good to me, thanks a lot again for all the help and support in our Adapter implementation and issue investigation. I really appreciate it. Have a great day! 😊

ericleich: Thanks for the review @YueVungle and for catching the above errors!

PR #233, opened 2020-05-20
Chartboost SDK 8.1.0 compatibility fixes v3

daniel-barros:
Fixes two issues:
- If framework set in GADMChartboostExtras is CBFrameworkUnity, `[Chartboost setFramework:withVersion:]` is called. This was not the case due to the `if (extras.framework)` check equating to NO if framework value is 0 (which corresponds to CBFrameworkUnity).

- Race condition happening due to access to GADMAdapterChartboostSingleton's `_completionHandlers` not being thread-safe. This array may be updated before the `[Chartboost startWithAppID:appSignature:completion]` call while on a different thread the same array is iterated over or cleared as part of the completion execution.
I fixed this last issue by removing the the whole `_completionHandlers `thing and relying on Chartboost SDK own handling of multiple initialization completion callbacks, which is thread-safe already.

Also substitutes Chartboost header imports by conditional imports to support Chartboost's testing tools.


adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
```diff
@@ -45,56 +36,20 @@ + (nonnull GADMAdapterChartboostSingleton *)sharedInstance {
...
-      break;
-    case GADMAdapterChartboostInitializing:
-      GADMAdapterChartboostMutableArrayAddObject(self->_completionHandlers, completionHandler);
-      break;
-    case GADMAdapterChartboostUninitialized:
-      GADMAdapterChartboostMutableArrayAddObject(self->_completionHandlers, completionHandler);
-      self->_initState = GADMAdapterChartboostInitializing;
-
-      GADMAdapterChartboostSingleton *weakSelf = self;
       [Chartboost startWithAppId:appId
```
joshuajplagonera, line 42: Can you confirm the following regarding the `startWithAppId:` initialization method?

* Can it be called more than once and the completion handler is called synchronously every time ?
* Will it return immediately if Chartboost SDK has already been initialized?

daniel-barros: Of course, glad to help.
- Yes it can be called more than once, even while Chartboost is still starting. Each completion handler will be called, one at a time, on the main thread.
- And yes, it will return immediately, on the main thread. This means that if the startWithAppId: method is called on a background thread, the caller function may return before the completion has been executed on the main thread.

joshuajplagonera: Thank you for clarifying!

joshuajplagonera:
Hi @daniel-barros 
Can you update this PR with the latest changes?

daniel-barros:
Hey @joshuajplagonera, it's done

ericleich:
Updating CLA because Joshua is from google and it was Joshua's commit that was failing.

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
```diff
@@ -45,56 +36,20 @@ + (nonnull GADMAdapterChartboostSingleton *)sharedInstance {
...
 
-- (nonnull instancetype)init {
-  self = [super init];
-  if (self) {
-    _completionHandlers = [[NSMutableArray alloc] init];
-  }
-  return self;
-}
-
 - (void)startWithAppId:(nonnull NSString *)appId
```
joshuajplagonera, line 39: This class now loses its use since we can call `[Chartboost startWithAppId:]` directly. 
Please delete this class and update the other classes to call `[Chartboost startWithAppId:]` directly.

adapters/Chartboost/ChartboostAdapter/GADMChartboostError.h
```diff
@@ -12,8 +12,16 @@
...
+#if __has_include(<Chartboost/Chartboost.h>)
 #import <Chartboost/Chartboost.h>
+#else
+#import "Chartboost.h"
+#endif
+#if __has_include(<Chartboost/CHBAdDelegate.h>)
+#import <Chartboost/CHBAdDelegate.h>
+#else
+#import "CHBAdDelegate.h"
+#endif
```
joshuajplagonera, line 24: Is it necessary to add a separate include to this file? 
Is this not included under `Chartboost.h` ?

adapters/Chartboost/ChartboostAdapter/GADMAdapterChartboostSingleton.m
daniel-barros: Makes sense 👍

adapters/Chartboost/ChartboostAdapter/GADMChartboostError.h
daniel-barros: Not necessary, I removed it.

daniel-barros:
Hi @joshuajplagonera, is there anything else we can do to move this forward?

PR #235, opened 2020-05-26
Facebook Rewarded Interstitial iOS Adapter

chlandacastro:
Facebook Rewarded Interstitial iOS Adapter


ericleich:
Hi @yroxe please review this when you get the chance!

PR #237, opened 2020-06-04
Removed IA logs

AviGelkop:
Temporary removed logs because our logger class reveals issue in FirebaseRemoteConfig

### DIFF
--- a/adapters/Fyber/FyberAdapter/GADMAdapterFyberUtils.m
+++ b/adapters/Fyber/FyberAdapter/GADMAdapterFyberUtils.m
@@ -104,7 +104,6 @@ BOOL GADMAdapterFyberInitializeWithAppID(NSString *_Nonnull appID,
     return NO;
   }
 
-  [IALogger setLogLevel:IALogLevelInfo];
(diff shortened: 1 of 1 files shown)


adapters/Fyber/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Fyber iOS Mediation Adapter Changelog
 
+#### Version 7.5.6.1
+- Verified compatibility with Fyber Marketplace SDK version 7.5.6.
+- Removed Fyber logs
```
joshuajplagonera, line 5: This might be misleading to think that all logging from the adapter has been removed.
Maybe revise to: "Temporarily removed Fyber Marketplace SDK-specific logging."

adapters/Fyber/CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 ## Fyber iOS Mediation Adapter Changelog
 
+#### Version 7.5.6.1
+- Verified compatibility with Fyber Marketplace SDK version 7.5.6.
```
joshuajplagonera, line 4: This can be removed since it is already included on the previous version.

PR #246, opened 2020-06-20
Vungle iOS AdMob Adapter 6.7.0.0 Release

YueVungle:
This version of adapter is compatible with Vungle iOS SDK 6.7.0.


YueVungle:
@googlebot I consent.

ashinagawa:
@googlebot I consent.

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
```diff
@@ -35,13 +33,20 @@ typedef NS_ENUM(NSUInteger, BannerRouterDelegateState) {
...
+- (void)willCloseAd;
+- (void)didCloseAd;
+- (void)trackClick;
 
 @optional
 
-// Get banner request object
-@property(nonatomic, nonnull) GADMAdapterVungleBannerRequest *bannerRequest;
+// Differentiate two banners with same placement ID
+@property(nonatomic, nullable) NSString *uniquePubRequestID;
```
ericleich, line 43: (nonatomic, copy, nullable)

Make it be a copy instead of being a strong reference, then other code calling this doesn't need to explicitly call copy when referring to this property.

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
```diff
@@ -35,13 +33,20 @@ typedef NS_ENUM(NSUInteger, BannerRouterDelegateState) {
...
+
+// Reward user for watching a Rewarded Video Ad successfully
+- (void)rewardUser;
+
+// App will leave current application
+- (void)willLeaveApplication;
```
ericleich, line 49: Can we make these required and have the classes conforming too this delegate just do nothing? That way we need less "if respondsToSelector" calls.

adapters/Vungle/CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 ## Vungle iOS Mediation Adapter Changelog
 
+#### Version 6.7.0.0
+- Verified compatibility with Vungle SDK 6.7.0.
+- Added support for playing multi ads at the same time.
+- Added support for new Vungle SDK event handlers.
```
ericleich, line 6: This should be more actionable for publishers. Other than being compatible with 6.7.0, I don't think publishers care about event handlers or init API.

Things that I think publishers will care more about:

- Clicks now reported when the click happens instead of when the ad is closed.
- Banner and interstitial ads now forward the `willLeaveApplication` callback.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -93,13 +95,29 @@ - (void)initWithAppId:(nonnull NSString *)appId
...
 
   // Disable refresh functionality for all banners
   [[VungleSDK sharedSDK] disableBannerRefresh];
 
+  // Set init options for priority placement
+  NSMutableDictionary *initOptions = [NSMutableDictionary dictionary];
+  if (delegate) {
+    NSString *priorityPlacementID = delegate.desiredPlacement;
+    [initOptions setObject:priorityPlacementID forKey:VungleSDKInitOptionKeyPriorityPlacementID];
+    _prioritizedPlacementID = [priorityPlacementID copy];
```
ericleich, line 107: Don't need copy if the property itself becomes copy.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -93,13 +95,29 @@ - (void)initWithAppId:(nonnull NSString *)appId
...
+
+    NSInteger priorityPlacementAdSize = 1;
```
ericleich, line 109: What is 1? Is that one of your ad size constants? If so, use the constant directly here for code clarity.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -93,13 +95,29 @@ - (void)initWithAppId:(nonnull NSString *)appId
...
+    if ([delegate respondsToSelector:@selector(bannerAdSize)]) {
+      GADAdSize adSize = [delegate bannerAdSize];
+      if (!GADAdSizeEqualToSize(adSize, kGADAdSizeMediumRectangle)) {
```
ericleich, line 112: Please add a comment here about why MREC is a special case.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -278,38 +262,24 @@ - (BOOL)hasDelegateForPlacementID:(nonnull NSString *)placementID {
...
-      return _bannerRequest == nil || [_bannerRequest isEqualToBannerRequest:bannerRequest];
-    }
-    return YES;
-  }
-}
-
 - (nullable NSError *)loadAd:(nonnull NSString *)placement
                 withDelegate:(nonnull id<GADMAdapterVungleDelegate>)delegate {
+  if (!delegate) {
+    return GADMAdapterVungleErrorWithCodeAndDescription(
```
ericleich, line 268: Can remove. This input is nonnull input and only adapter classes call this method. This code only masks an adapter bug that's not respecting nonnull, whic h

If you found a reason why this needs to be here, please fix the caller to not pass in a nil delegate.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -318,6 +288,12 @@ - (nullable NSError *)loadAd:(nonnull NSString *)placement
     return nil;
   }
 
+  // We already requested an ad for _prioritizedPlacementID,
+  // so we don't need to request again.
```
ericleich, line 292: Is thiis really ok? I see you pass this ID into init, but that doesn't necessarily mean an ad is cached. In fact, we already confirmed in the code above that the ad ISN'T cached. Shouldn't we still call loadPlacementWithID: and wait for a callback? What happens during a refresh or the next time an ad is requested for this placement?

Put another way, is there any reason we can't just fall through to the default behavior and call loadPlacementWithID: again? I'd prefer to get rid of the _prioritizedPlacementID in this class if we don't need it. I would assume Vungle SDK can handle an additional loadPlacementWithID: call.

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
YueVungle: @ericleich Hi Eric, I'm sorry for the late response. I took PTO yesterday. Thanks a lot for the guidance, I will add 'copy' attribute.  :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
YueVungle: @ericleich Hi Eric, you mean we make `- (void)willLeaveApplication` as required not for `- (void)rewardUser`, right? Because I think `- (void)rewardUser` is only for Rewarded Video ad ? Thanks a lot. ^_^

adapters/Vungle/CHANGELOG.md
YueVungle: @ericleich Hi Eric, what you said makes much sense. I will remove `Added support for new Vungle SDK event handlers` and `Added support for new Vungle SDK init API`, then I will add your suggestions. Thanks for the instructions. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, we only defined an instance variable `_prioritizedPlacementID`, we didn't define a property `prioritizedPlacementID`, so we add copy method here. Thanks a lot. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich  Hi Eric, yes 1 is for `VungleAdSizeUnknown` enum. I will use `VungleAdSizeUnknown` directly here. Thanks for the guidance. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich  Got it, thanks. :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Got it, I will remove this non nil check. Thanks a lot for the instruction. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: Hi Eric, I'm really sorry for the confusing, the `_prioritizedPlacementID` logic is a workaround of a bug for Vungle SDK sequential download with prioritized  placement:

If we add a  placementID in a dictionary then we pass this dictionary to the `options` parameter of new Vungle SDK init api: `- (BOOL)startWithAppId:options:error:`, Vungle SDK will initialize and begin to load this placementID. Then if we call `loadPlacementWithID:` method to load this placementID which is already beginning to load. The download process will stuck. 

So we add this `_prioritizedPlacementID` related logic, if a placementID is already beginning to load in the new Vungle SDK init api: `- (BOOL)startWithAppId:options:error:`, we will not load it again in the adapter to avoid stuck the download process. After the SDK initialization is finished,  the ` _prioritizedPlacementID = nil;`, so the refresh or the next time an ad is requested for this placement can work. 

We will remove this `_prioritizedPlacementID` logic in the next version after the stuck issue is fixed. Thanks so much. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
YueVungle: @ericleich Added, thanks. ^_^

adapters/Vungle/CHANGELOG.md
YueVungle: @ericleich Updated, thanks. 🙂

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Updated, thanks a lot. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Added, thanks. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Removed, thanks and have a great day! 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
ericleich: @YueVungle I'd like to make both required for every adapter to implement. It's a side effect of every adapter type using the same GADMAdapterVungleDelegate. Agree that banner/interstitial will have an empty implementation for `- (void)rewardUser` but I'd prefer that as it's more explicit that the adapter is taking no action. It also enforces that rewarded adapters have to implement rewardUser, for example, instead of omitting it. I feel the more explicit no-op avoids future bugs. Additionally, we can then get rid of `if respondsToSelector:` checks.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: Ah, so `@property(nonatomic, nonnull) NSString *desiredPlacement;` on GADMAdapterVungleDelegate should also be copy. That way, way, you can just write:

```
_prioritizedPlacementID = delegate.desiredPlacement
```

And you will get a copy by default as you're accessing the desiredPlacement property.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -93,13 +95,29 @@ - (void)initWithAppId:(nonnull NSString *)appId
...
+    _prioritizedPlacementID = [priorityPlacementID copy];
+
+    NSInteger priorityPlacementAdSize = 1;
+    if ([delegate respondsToSelector:@selector(bannerAdSize)]) {
+      GADAdSize adSize = [delegate bannerAdSize];
+      if (!GADAdSizeEqualToSize(adSize, kGADAdSizeMediumRectangle)) {
+        priorityPlacementAdSize = GADMAdapterVungleAdSizeForCGSize(adSize.size);
+      }
+    }
+    [initOptions setObject:[NSNumber numberWithInteger:priorityPlacementAdSize] forKey:VungleSDKInitOptionKeyPriorityPlacementAdSize];
```
ericleich, line 119: @YueVungle I see here you set init options to "unknown" for the size in the MREC case. is that ok?

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: I see. Can you update the comment to say something like:

// Vungle 6.7.0 SDK cannot handle a second `loadPlacementWithID:` call while the first ad load is in progress. Work around this behavior by explicitly avoiding calling `loadPlacementWithID:`on the first request for this placement, as the priority placement has already started loading at initialization time.

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
YueVungle: @ericleich Hi Eric, thanks a lot for the detailed explanation, I will make `- (void)willLeaveApplication` and `- (void)rewardUser` as required. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, I have a concern that the `copy` attribute is only for setter method, not for getter method:

Excerpt from stackoverflow -----
The `copy` attribute means only that the setter will take a copy when the property is set. The synthesized getter will not return a copy. You must implement that behavior yourself:
https://stackoverflow.com/questions/25111218/do-i-need-to-override-the-getter-for-a-property-to-return-an-immutable-copy

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, got it, thanks a lot for the guidance, I will update the comment.  :)

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, yes, only for Banner ad type (not include MREC) needs the Banner ad size.  Thanks 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleDelegate.h
YueVungle: @ericleich Updated, thanks. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Updated, thanks and have a great day. O(∩_∩)O

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: Hmm, but I think NSString behaves differently is different?

I ran this test:

```
  NSString *a = @"a";
  NSString *other = a;
  a = @"b";
  NSLog(@"Other string: %@", other);
```

and it logged;

```
Other string: a
```

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
```diff
@@ -213,4 +213,8 @@ - (void)willLeaveApplication {
   [_connector adapterWillLeaveApplication:_adapter];
 }
 
+- (void)rewardUser {
+  NSLog(@"%@ received rewardUser event.", NSStringFromClass(self.class));
```
ericleich, line 217: This can just be:

```
// Do nothing.
```

I don't want to confuse users if this log were to make it through.

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -180,4 +180,8 @@ - (void)willLeaveApplication {
   [_connector adapterWillLeaveApplication:self];
 }
 
+- (void)rewardUser {
+  NSLog(@"%@ received rewardUser event.", NSStringFromClass(self.class));
```
ericleich, line 184: Same thing for these other events that do nothing.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
ericleich: Yeah, NSString is special. I also ran this test:

```
  GADInterstitial *interstitial = [[GADInterstitial alloc] initWithAdUnitID:@"abc"];
  GADInterstitial *otherInterstitial = interstitial;
  interstitial.adUnitID = @"def";
  NSLog(@"Other interstitial ad unit: %@", otherInterstitial.adUnitID);
```

and got:

```
Other interstitial ad unit: def
```

So while pointers are typically soft copy, NSString seems to make a deep copy on get.

ericleich: If you feel strongly about this, I can accept. But it's quite possible we may do an internal refactor later. I haven't seen code needing to copy strings (other than setter property being copy), and if that were the case, I feel like I'd see it everywhere.

As another example, your code doesn't pass a copy to initOptions. Wouldn't you want to make a copy there if we were doing copies of strings?

ericleich:
@YueVungle looking pretty good, just a couple more touch up comments. Apologies for the slow turnaround, I'll try to check back in tomorrow to get this merged.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
YueVungle: @ericleich Hi Eric, thanks so much for showing me the code snippets,  I will update the code according to your instructions. ^_^

YueVungle: @ericleich Updated, thanks a lot. 😊

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
YueVungle: @ericleich Updated, thanks a lot for the guidance. ^_^

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
YueVungle: @ericleich Updated, thanks a lot for the guidance. ^_^

YueVungle:
> @YueVungle looking pretty good, just a couple more touch up comments. Apologies for the slow turnaround, I'll try to check back in tomorrow to get this merged.

@ericleich Hi Eric, no worries, I know you have a lot of work. I updated the adapter according to your suggestions. Many thanks for all the guidances and help in the adapter implementations. Have a great day! O(∩_∩)O

ericleich:
Thanks for the changes @YueVungle ! We'll be merging and putting out a release soon.

PR #259, opened 2020-08-28
AppLovin/remove_mrec_and_native_ads

tiandrew:


ericleich:
@tiandrew Thanks for the PR. At first glance, changes look good. We'll do a more thorough review and testing soon so we can merge this in.

I did have a general question though about how banner sizes are loaded. In https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovin.m#L180-L184 we call:

```
[_sdk.adService loadNextAdForZoneIdentifier:_zoneIdentifier andNotify:_bannerDelegate];
```

In situations where there's a known zone identifier. In practice, we expect there to always be a zone ID since we ask for it in the AdMob UI when configuring AppLovin for mediation. But the above API does not have the ad size associated with it. How does AppLovin know which size to load in this situation (320x50 vs 728x90)? In the AppLovin UI, when a zone is created, I only see a way to specify "Banner", but not a 320x50 vs 728x90 size. So I don't understand where you get that information from.

adapters/AppLovin/AppLovinAdapter/GADMediationAdapterAppLovin.m
```diff
@@ -121,14 +121,6 @@ - (void)collectSignalsForRequestParameters:(nonnull GADRTBRequestParameters *)pa
                          completionHandler:
                              (nonnull GADRTBSignalCompletionHandler)completionHandler {
   [GADMAdapterAppLovinUtils log:@"AppLovin adapter collecting signals."];
-  // Check if supported ad format.
```
santoshbagadi, line 124: @tiandrew , we should keep this check in right? We are essentially making sure we don't return a signal for Native.

tiandrew:
> @tiandrew Thanks for the PR. At first glance, changes look good. We'll do a more thorough review and testing soon so we can merge this in.
> 
> I did have a general question though about how banner sizes are loaded. In https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/adapters/AppLovin/AppLovinAdapter/GADMAdapterAppLovin.m#L180-L184 we call:
> 
> ```
> [_sdk.adService loadNextAdForZoneIdentifier:_zoneIdentifier andNotify:_bannerDelegate];
> ```
> 
> In situations where there's a known zone identifier. In practice, we expect there to always be a zone ID since we ask for it in the AdMob UI when configuring AppLovin for mediation. But the above API does not have the ad size associated with it. How does AppLovin know which size to load in this situation (320x50 vs 728x90)? In the AppLovin UI, when a zone is created, I only see a way to specify "Banner", but not a 320x50 vs 728x90 size. So I don't understand where you get that information from.

We check if it's a banner or leader by using Apple's provided API: `[[UIDevice currentDevice] userInterfaceIdiom] == UIUserInterfaceIdiomPad`, for iPads, we use the larger 728x90, and if not we'll use the 320x50

PR #270, opened 2020-10-05
Unity Adapter update to UnityAds SDK v3.5.0

jindoujiao:
Unity Adapter updates for UnityAdsSDK v3.5.0.

- Leverage new IUnityAdsInitializationListener get UnityAds Initialization callbacks.
- Leverage new IUnityAdsLoadListener to get UnityAds load() callbacks.
- Limited the usage of original IUnityAdsExtendedListener for UnityAds show() callbacks only.
- Created dedicate classes for RewardedAd and BannerAd.
- Bug fixes.


jindoujiao:
@googlebot I consent.

adapters/Unity/UnityAdapter/GADMAdapterUnityBannerAd.m
```diff
@@ -49,27 +52,36 @@ - (nonnull instancetype)initWithGADMAdNetworkConnector:(nonnull id<GADMAdNetwork
...
+    NSError *error = GADMAdapterUnityErrorWithCodeAndDescription(
+           GADMAdapterUnityErrorInvalidServerParameters, @"Game ID and Placement ID cannot be nil.");
+    [strongConnector adapter:strongAdapter didFailAd:error];
     return;
   }
 
   if (![UnityAds isInitialized]) {
-    NSString *gameID = [strongConnector.credentials[kGADMAdapterUnityGameID] copy];
-    [GADMAdapterUnitySingleton.sharedInstance initializeWithGameID:gameID];
+    [[GADMAdapterUnity alloc] initializeWithGameID:_gameID withInitDelegate:Nil];
```
joshuajplagonera, line 76: I would assume that the banner ad will fail to load or initialize if the UnityAds fail to initialize here?

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -83,100 +63,61 @@ - (void)requestRewardedAd {
...
+     NSStringFromClass([UnityAds class])];
     if (_adLoadCompletionHandler) {
-      NSError *error = GADMAdapterUnityErrorWithCodeAndDescription(
-          GADMAdapterUnityErrorDeviceNotSupported, description);
+      NSError *error = GADMAdapterUnityErrorWithCodeAndDescription(GADMAdapterUnityErrorDeviceNotSupported, description);
       _adLoadCompletionHandler(nil, error);
       _adLoadCompletionHandler = nil;
     }
     return;
   }
```
joshuajplagonera, line 74: I think we can remove this part since `initializeWithGameID` already checks this.

adapters/Unity/UnityAdapter/GADMUnityInitializer.m
```diff
@@ -0,0 +1,189 @@
...
+
+- (void)stopBeingDelegate {
+  if (_bannerAd != nil) {
+    [_bannerAd stopBeingDelegate];
+  }
+}
+
+- (void)initializeWithGameID:(NSString *)gameID withInitDelegate:(id)initDelegate{
+  if (![UnityAds isSupported]) {
+    NSLog(@"Unity Ads cannot be initialized: this device is not supported.");
```
joshuajplagonera, line 65: `return;` after here.

adapters/Unity/UnityAdapter/GADMUnityInitializer.m
```diff
@@ -0,0 +1,189 @@
...
+  }
+  
+  if ([UnityAds isInitialized]) {
+    NSLog(@"Unity Ads has already been initialized.");
```
joshuajplagonera, line 69: `return;` after this line

joshuajplagonera (changes requested):
First round of review.
We are having issues with the UnityAds 3.5.0 iOS SDK. When trying to import it into the `Drop_Headers_And_Frameworks` folder, the UnityAds SDK cannot be found.
This only happens to 3.5.0 SDK. 

3.4.8 SDK is detected. Both xCode 11 and xCode 12 sees this issue.
Can you take a look?

adapters/Unity/UnityAdapter/GADMAdapterUnityBannerAd.m
jindoujiao: Yes, if UnityAds failed to initialize banner load call will fail.

joshuajplagonera:
Hi @jindoujiao ,

We are having issues with the UnityAds 3.5.0 iOS SDK. When trying to import it into the `Drop_Headers_And_Frameworks` folder, the UnityAds SDK cannot be found.
This only happens to 3.5.0 SDK.

3.4.8 SDK is detected. Both xCode 11 and xCode 12 sees this issue.
Can you take a look?

jindoujiao:
> Hi @jindoujiao ,
> 
> We are having issues with the UnityAds 3.5.0 iOS SDK. When trying to import it into the `Drop_Headers_And_Frameworks` folder, the UnityAds SDK cannot be found.
> This only happens to 3.5.0 SDK.
> 
> 3.4.8 SDK is detected. Both xCode 11 and xCode 12 sees this issue.
> Can you take a look?

Hi @joshuajplagonera , 

I'm working with the team and investigate the issue. 
Are you integrate with cocoapod via ` pod 'UnityAds', '3.5.0'`?

joshuajplagonera:
> > Hi @jindoujiao ,
> > We are having issues with the UnityAds 3.5.0 iOS SDK. When trying to import it into the `Drop_Headers_And_Frameworks` folder, the UnityAds SDK cannot be found.
> > This only happens to 3.5.0 SDK.
> > 3.4.8 SDK is detected. Both xCode 11 and xCode 12 sees this issue.
> > Can you take a look?
> 
> Hi @joshuajplagonera ,
> 
> I'm working with the team and investigate the issue.
> Are you integrate with cocoapod via ` pod 'UnityAds', '3.5.0'`?

HI @jindoujiao ,

I am still trying to build the adapter and not including it in a sample app yet. But I am not able to build the adapter since it cannot find the UnityAds SDK when included in the `Drop_Framework_and_Headers` folder.
This issue does not occur for UnityAds version 3.4.8 though.

jindoujiao:
@joshuajplagonera Could you please re-download the framework from this link?
 https://github.com/Unity-Technologies/unity-ads-ios/releases/tag/3.5.0

joshuajplagonera:
Hi @jindoujiao ,

Yes, that is where I downloaded the UnityAds.framework file to include to the `Drop_Framework_And_Headers` folder. However, I tried re-downloading it from there and it doesn't seem to help. Any ideas?

joshuajplagonera:
Hi @jindoujiao 
Thanks for the assistance. It looks like I was not able to import it properly.
I will go ahead with the review now and provide additional comments where applicable.

jindoujiao:
@joshuajplagonera we changed to  package the framework bundled with the license since 3.5.0. 
Could you try to unzip and extract the zip and copy the framework inside?

adapters/Unity/UnityAdapter/GADMUnityInitializer.m
```diff
@@ -0,0 +1,195 @@
...
+    return;
+  }
+  [_interstitialAd getInterstitial];
+}
+
+- (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewController {
+  id<GADMAdNetworkConnector> strongConnector = _networkConnector;
+  if (strongConnector) {
+    [strongConnector adapterWillPresentInterstitial:self];
+  }
```
joshuajplagonera, line 116: Remove. This is causing multiple `willPresentInterstitial` callbacks when trying to present an interstitial ad.
Additionally, `_interstitialAd` is already forwarding this callback.

joshuajplagonera (changes requested):
Hi @jindoujiao ,

I believe we should retain the `GADMAdapterUnity` filename instead of using `GADMUnityInitializer`. Can you revert the filename?

jindoujiao:
> Hi @jindoujiao ,
> 
> I believe we should retain the `GADMAdapterUnity` filename instead of using `GADMUnityInitializer`. Can you revert the filename?

I reverted the `GADMUnityInitializer` back to `GADMAdapterUnity`. I prefer to revert it back to track the file changes. 
Thanks for the suggestion.

ericleich:
@richardh-unity3d @AliceNoWonderland Can you acknowledge in the comments that you're ok with your commits being committed to this repo under https://cla.developers.google.com/clas? If you both acknowledege, I can manually flip the CLA check.

ericleich:
@richardh-unity3d @AliceNoWonderland  Looks like `@googlebot I consent.` should do it.

AliceNoWonderland:
@googlebot I consent.

ericleich:
Manually overriding CLA as @richardh-unity3d is on leave.

ericleich:
@jindoujiao I just completed an internal review on this CL (and will have Joshua follow up with more comments based on that feedback), but a couple of high level questions:

1) Can Unity now load guarantee that X different load requests for that same placement result in X unique ads? The adapter no longer checks for multiple requests for the same placement.
2) Can Unity handle multiple init calls at once before the first call finishes? If I call init 5 times back to back with 5 different delegate implementations, will you call back all 5 delegates once init finishes?
3) Can Unity handle a loadAd call before initialization is complete? The adapter does not wait for the init callback before loading ads.
4) If init fails, should we immediately fail the ad request? In other words, can Unity SDK recover from an init failure? The code will just call init and then call load immediately afterwards if the adapter is not already initialized. Since the load failure callback has no info about why it failed, the best we can say is "no fill", but that may be masked by an init failure in the current implementation.

AliceNoWonderland:
> @jindoujiao I just completed an internal review on this CL (and will have Joshua follow up with more comments based on that feedback), but a couple of high level questions:
> 
> 1. Can Unity now load guarantee that X different load requests for that same placement result in X unique ads? The adapter no longer checks for multiple requests for the same placement.
> 2. Can Unity handle multiple init calls at once before the first call finishes? If I call init 5 times back to back with 5 different delegate implementations, will you call back all 5 delegates once init finishes?
> 3. Can Unity handle a loadAd call before initialization is complete? The adapter does not wait for the init callback before loading ads.
> 4. If init fails, should we immediately fail the ad request? In other words, can Unity SDK recover from an init failure? The code will just call init and then call load immediately afterwards if the adapter is not already initialized. Since the load failure callback has no info about why it failed, the best we can say is "no fill", but that may be masked by an init failure in the current implementation.

@ericleich hope the following can answer your questions
1. No, X different load requests will not result in X unique ads, but will get X different load callbacks. The reason we don't need to check for multiple requests for the same placement is the load callback will take care of each load request
2, Yes, we can and that 5 delegates call will triggered once init finishes
3. Yes, if a load call is triggered before initialization is finished, Unity sdk will wait till it is done initialization and then send that load call automatically
4.There might be a small chance we can recover from init failure. I think 'no fill' for this is okay to us.

ericleich:
@AliceNoWonderland For #1, similar to Android, we need the check in place to ensure 2 loads for the same placement aren't backed by what might be the same Unity ad.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -0,0 +1,194 @@
...
+  BOOL _isLoading;
+  /// Connector from Google Mobile Ads SDK to receive ad configurations.
+  __weak id<GADMAdNetworkConnector> _connector;
+  
+  /// Adapter for receiving ad request notifications.
+  __weak id<GADMAdNetworkAdapter> _adapter;
+}
+
+  /// A dictionary to keep track loaded Placement IDs
+  static NSMutableDictionary<NSString *, id> *_placementInUse;
```
ericleich, line 36: add type. Also consider NSMapTable

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -0,0 +1,194 @@
...
+           GADMAdapterUnityErrorInvalidServerParameters, @"Game ID and Placement ID cannot be nil.");
+    [strongConnector adapter:strongAdapter didFailAd:error];
+    return;
+  }
+
+  if (![UnityAds isInitialized]) {
+    [[GADMAdapterUnity alloc] initializeWithGameID: _gameID withInitDelegate:Nil];
+  }
+
+  @synchronized (_placementInUse) {
```
ericleich, line 78: libDispatch instead of synchronized.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -0,0 +1,194 @@
...
+    NSLog(@"Unity Ads Adapter Error: No GADMAdNetworkConnector found.");
+    return;
+  }
+  
+  if (!strongAdapter) {
+    NSLog(@"Unity Ads Adapter Error: No GADMAdNetworkAdapter found.");
+    return;
+  }
+  
+  if ([UnityAds isReady:_placementID]) {
```
ericleich, line 111: if (![UnityAds isReady:_placementID]) {
    NSError *error = GADMAdapterUnityErrorWithCodeAndDescription(GADMAdapterUnityErrorShowAdNotReady, @"Failed to show Unity Ads interstitial.");
    [_adEventDelegate didFailToPresentWithError:error];
    return;
  }

adapters/Unity/UnityAdapter/GADMediationAdapterUnity.h
```diff
@@ -36,9 +36,12 @@ typedef NS_ENUM(NSInteger, GADMAdapterUnityErrorCode) {
...
   GADMAdapterUnityErrorAdAlreadyLoaded = 108,
   /// Banner size mismatch.
-  GADMAdapterUnityErrorSizeMismatch = 109
+  GADMAdapterUnityErrorSizeMismatch = 109,
+  /// UnityAds returned an initialization error
+  GADMAdapterUnityErrorAdInitializationFailure = 110
 };
 
 @interface GADMediationAdapterUnity : NSObject <GADMediationAdapter>
+@property (nonatomic, copy) GADMediationAdapterSetUpCompletionBlock _Nonnull completionBlock;
```
ericleich, line 45: Remove this? it doesn't look like it's used in the .m

adapters/Unity/UnityAdapter/GADMediationAdapterUnity.m
```diff
@@ -51,9 +49,8 @@ + (void)setUpWithConfiguration:(GADMediationServerConfiguration *)configuration
           gameIDs);
     NSLog(@"Initializing Unity Ads SDK with the game ID %@.", gameID);
   }
-
-  [[GADMAdapterUnitySingleton sharedInstance] initializeWithGameID:gameID];
-  completionHandler(nil);
+  GADMUnityInitializationDelegate* initializationDelegate = [[GADMUnityInitializationDelegate alloc] initWithCompletionHandler:completionHandler];
```
ericleich, line 52: Per other comments earlier, it'd be nice if this class can just forward on the block, and expect GADMAdapterUnity to call it at the right time. How GADMAdapterUnity does that is an implementation detail that this class shouldn't need to worry about.

ericleich (changes requested):

ericleich:
@AliceNoWonderland Please take a look.

AliceNoWonderland:
> @AliceNoWonderland Please take a look.

@ericleich thank you so much for the detailed review and sorry that I missed these comments last week. I believe all the comments are addressed now :) Please let us know if there is any other changes we need to make

adapters/Unity/UnityAdapter/GADMAdapterUnity.m
```diff
@@ -13,33 +13,42 @@
...
 
-  /// MetaData for storing Unity instrument analysis
-  UADSMetaData *_metaData;
+@interface GADMUnityInitializationDelegate : NSObject {
+  GADMediationAdapterSetUpCompletionBlock initCompletionBlock;
 }
 
+- (nonnull instancetype)initWithCompletionHandler:(GADMediationAdapterSetUpCompletionBlock)completionHandler;
+
+- (void)initializationComplete;
```
ericleich, line 49: Shouldn't need to define this API here. This is just Unity's UnityAdsInitializationDelegate API, which is defined by saying this class implements that delegate.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -67,27 +70,27 @@ - (void)requestRewardedAd {
...
   
   if (![UnityAds isInitialized]) {
-    [[GADMAdapterUnity alloc] initializeWithGameID:_gameID withInitDelegate:Nil];
+    [[GADMAdapterUnity alloc] initializeWithGameID:_gameID withCompletionHandler:nil];
   }
 
-  @synchronized (_placementInUseRewarded) {
+  dispatch_sync(_lockQueue, ^{
```
ericleich, line 76: This sync is too long. If the completion handler invokes code that does another sync on this _lockQueue, then we're blocked.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -67,27 +70,27 @@ - (void)requestRewardedAd {
...
+  });
 
   [UnityAds load:_placementID loadDelegate:self];
 }
 
 - (void)presentFromViewController:(nonnull UIViewController *)viewController {
-  @synchronized (_placementInUseRewarded) {
-    [_placementInUseRewarded removeObjectForKey:_placementID];
-  }
+  dispatch_sync(_lockQueue, ^{
```
ericleich, line 91: Per https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/CONTRIBUTING.md#best-practices, writes can be asynchronous. Only reads need to be synchronous, which will block if any writes haven't been written.

ericleich (approved):
@AliceNoWonderland I have a few remaining comments but it'll probably be easier for us to clean these up as part of doing the release (we will do things like run our formatter anyways). Will merge after @joshuajplagonera signs off.

ericleich:
@AliceNoWonderland Also thank you for making the previous requested changes!

AliceNoWonderland:
@ericleich thank you again for the detailed reviews, I guess we will rely on you guys for the remaining changes ? Please let me know if you need anything else from us!

PR #274, opened 2020-10-20
Updating myTarget iOS Adapter to version 5.9.2.0

aserrr:
Supported GoogleMobileAds 7.66.0 and myTarget 5.9.2

### DIFF
--- a/adapters/MyTarget/MyTargetAdapter/GADMAdapterMyTarget.m
+++ b/adapters/MyTarget/MyTargetAdapter/GADMAdapterMyTarget.m
@@ -72,7 +72,6 @@ static GADAdSize GADSupportedAdSizeFromRequestedSize(GADAdSize gadAdSize) {
 
 - (void)getBannerWithSize:(GADAdSize)adSize {
   MTRGLogInfo();
-  MTRGLogDebug(@"adSize: %.fx%.f", adSize.size.width, adSize.size.height);
 
   id<GADMAdNetworkConnector> strongConnector = _connector;
   if (!strongConnector) {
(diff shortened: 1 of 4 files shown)


adapters/MyTarget/MyTargetAdapter/GADMAdapterMyTarget.m
```diff
@@ -72,21 +72,23 @@ - (nonnull instancetype)initWithGADMAdNetworkConnector:
...
   if (GADAdSizeEqualToSize(adSize, kGADAdSizeBanner)) {
-    adViewSize = MTRGAdSize_320x50;
+    adViewSize = [MTRGAdSize adSize320x50];
   } else if (GADAdSizeEqualToSize(adSize, kGADAdSizeMediumRectangle)) {
-    adViewSize = MTRGAdSize_300x250;
+    adViewSize = [MTRGAdSize adSize300x250];
   } else if (GADAdSizeEqualToSize(adSize, kGADAdSizeLeaderboard)) {
-    adViewSize = MTRGAdSize_728x90;
+    adViewSize = [MTRGAdSize adSize728x90];
+  } else if (adSize.size.width < 0 && adSize.size.height < 0) {
```
ericleich, line 89: instead of overriding here when kGADAdSizeInvalid is returned here, I suggest you add 

[MTRGAdSize adSizeForCurrentOrientationForWidth:adSize.size.width] as a supported ad size in the GADSupportedAdSizeFromRequestedSize() function, and let the GMA SDK function determine if that size is appropriate. We aren't sure what height your helper method returns, and we want to make sure it fits the incoming ad size (which GADSupportedAdSizeFromRequestedSize will check).

Or, are you ok to just set width and height of the _adView to the same size that is passed into the request, and see if your SDK can fill?

Keep in mind that the request coming in could theoretically be any size. Some publshers will use Google's adapter banner helper methods to create a size, but a publisher could create a custom ad size 360x140 (hypothetical example). We prefer either you pass 360x140 to your SDK to try to load, or pass [MTRGAdSize adSizeForCurrentOrientationForWidth:360] to the helper method so we can validate that the height you request is between 70-100% of the 140 passed in.

aserrr: Ok. Just added our size to the GADSupportedAdSizeFromRequestedSize() function.

PR #283, opened 2020-12-01
Add completion handler null check to prevent crash

AliceNoWonderland:
### Description
This pr is to add back the if check for completion handler which we previously added in the [pr](https://github.com/googleads/googleads-mobile-ios-mediation/pull/270/files#diff-3049d5d2e2115b254321aef39709c848cccde2e9956eef921f4776817cc68cefR195). 
As this completion handler could be null in many cases, without the check it causes app crash for many users.


AliceNoWonderland:
Thank you so much for the quick turnaround @ericleich

PR #286, opened 2021-01-04
Updated the Fyber Marketplace iOS adapter to use their new initialization APIs.

joshuajplagonera:
Updated the Fyber Marketplace iOS adapter to use their new initialization APIs.

### DIFF
--- a/adapters/Fyber/FyberAdapter/GADMAdapterFyberBannerAd.m
+++ b/adapters/Fyber/FyberAdapter/GADMAdapterFyberBannerAd.m
@@ -37,6 +37,12 @@
   /// GMA SDK, not set on the GMA SDK.
   id<GADMediationBannerAdEventDelegate> _delegate;
 
+  /// Fyber Ad Spot to be loaded.
+  IAAdSpot *_adSpot;
+
+  /// Fyber MRAID controller to support HTML ads.
+  IAMRAIDContentController *_MRAIDContentController;
+
   /// Fyber view controller to catch banner related ad events.
   IAViewUnitController *_viewUnitController;
 }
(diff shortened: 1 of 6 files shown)


adapters/Fyber/FyberAdapter/GADMAdapterFyberInitializer.h
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#import <Foundation/Foundation.h>
+
+/// Fyber adapter initialization completion handler.
+typedef void (^GADMAdapterFyberInitCompletionHandler)(NSError *_Nullable error);
+
+@interface GADMAdapterFyberInitializer : NSObject
```
joshuajplagonera, line 20: How does Fyber Marketplace SDK handle init in these situations:
1. Multiple simultaneous calls to init.
2. Called using different app IDs.

AviGelkop: 1. In a case where the SDK is in the process of the init, we will ignore the next calls up until we will finish with the init.
2. If the SDK finished with the init and it gets a new init request, the SDK will check if there is something that needs to be updated, such as a new app ID. In this case, the SDK will fetch the relevant configuration.

adapters/Fyber/FyberAdapter/GADMAdapterFyberInitializer.m
```diff
@@ -0,0 +1,100 @@
...
+  GADMAdapterFyberInitializer *__weak weakSelf = self;
+  [IASDKCore.sharedInstance
+        initWithAppID:appID
+      completionBlock:^(BOOL success, NSError *_Nullable error) {
+        GADMAdapterFyberInitializer *strongSelf = weakSelf;
+        if (!strongSelf) {
+          return;
+        }
+
+        if (error) {
```
AviGelkop, line 85: @joshuajplagonera did you mean `if (success)`?

AviGelkop (changes requested):

adapters/Fyber/FyberAdapter/GADMAdapterFyberInitializer.h
joshuajplagonera: In case of (2), how will the Fyber Marketplace SDK handle it if a different App ID was used to initialize the SDK a second time? Will it overwrite the existing configuration?

adapters/Fyber/FyberAdapter/GADMAdapterFyberInitializer.m
joshuajplagonera: Fixed. Thank you !

adapters/Fyber/FyberAdapter/GADMAdapterFyberInitializer.h
AviGelkop: Yes, in this case, the SDK will use the new configuration only. It means that if the spotID's from the previous appID won't be valid anymore and the request will fail.

joshuajplagonera: We updated the PR to let the Fyber SDK handle the initialization.
Please take a look

AviGelkop (approved):

PR #290, opened 2021-01-26
Release 6.9.1.0 Vungle AdMob Adapter

YueVungle:
This version of the adapter has been certified with Vungle 6.9.1 SDK.

### DIFF
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
@@ -168,6 +168,10 @@
 }
 
 - (void)willShowAd {
+  // Do nothing.
+}
+
+- (void)didViewAd {
   [_connector adapterWillPresentInterstitial:self];
 }
 
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
@@ -168,11 +168,11 @@
 }
 
 - (void)willShowAd {
-  // Do nothing.
+  [_connector adapterWillPresentInterstitial:self];
 }
 
 - (void)didViewAd {
-  [_connector adapterWillPresentInterstitial:self];
+  // Do nothing.
 }
 
 - (void)willCloseAd {
--- a/adapters/Vungle/CHANGELOG.md
+++ b/adapters/Vungle/CHANGELOG.md
@@ -3,6 +3,10 @@
 #### Next Version
 - Added standardized adapter error codes and messages.
 
+#### Version 6.9.0.0
+- Verified compatibility with Vungle SDK 6.9.0.
+- Now requires Google Mobile Ads SDK version 7.68.0 or higher.
+
 #### Version 6.8.1.0
 - Verified compatibility with Vungle SDK 6.8.1.
 - Now requires Google Mobile Ads SDK version 7.66.0 or higher.
--- a/adapters/Vungle/CHANGELOG.md
+++ b/adapters/Vungle/CHANGELOG.md
@@ -3,8 +3,8 @@
 #### Next Version
 - Added standardized adapter error codes and messages.
 
-#### Version 6.9.0.0
-- Verified compatibility with Vungle SDK 6.9.0.
+#### Version 6.9.1.0
+- Verified compatibility with Vungle SDK 6.9.1.
 - Now requires Google Mobile Ads SDK version 7.68.0 or higher.
 - Remove VungleSDKResetPlacementForDifferentAdSize error check for loading Ads.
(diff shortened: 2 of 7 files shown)


adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -168,6 +168,10 @@ - (void)adNotAvailable:(nonnull NSError *)error {
 }
 
 - (void)willShowAd {
+  // Do nothing.
+}
+
+- (void)didViewAd {
   [_connector adapterWillPresentInterstitial:self];
```
chlandacastro, line 175: Doesn't  it make more sense that `willShowAd` will invoke `willPresentInterstitial` ? shouldn't this go in willShowAd?

YueVungle: @chlandacastro Hi chlandacastro, based on our engineer @ashinagawa had the conversation with AdMob side(you might not be in that conversation) contacts before, the following was identified :
For Interstitial Ads:
The new SDK callback `vungleAdViewedForPlacement:` should be mapped to the `adapterWillPresentInterstitial:` adapter callback (which would also need to be removed from the `willShowAd` callback).
Thanks a lot. O(∩_∩)O

ericleich: Hey @YueVungle , was that conversation with me? If so, I don't remember it.

`adapterWillPresentInterstitial` is a callback to tell the user "an ad is about to take over the full screen". The Vungle router maps `vungleWillShowAdForPlacementID` to `willShowAd` and `vungleAdViewedForPlacement` to `didViewAd`. The meaning of Google's `adapterWillPresentInterstitial` callback sounds much closer to `vungleWillShowAdForPlacementID` like how it was before.

YueVungle: Hi @ericleich and @chlandacastro, `vungleAdViewedForPlacement:` is more for “impression timing (= at the very first frame is rendered)“, so we linked it to `adapterWillPresentInterstitial:`. 

My leader is taking PTO  and will come back to work tomorrow, after he comes back to work, I will confirm with him who was in that conversation, thanks a lot and have a great day! ^_^

crbinsf:
@ericleich, @chlandacastro - `didViewAd` was added to align more closely with `vungleAdViewedForPlacement:` to allow for more precise impression tracking (for rewarded ads). For intersitial and banner, if impressions are counted with the `adapterWillPresentInterstitial` callback, doesn’t it make sense to have them called from `didViewAd`?  Eric, during our original chat on this, it seems like that was the decision at the time.

ericleich:
@crbinsf While we do happen to record a mediation impression at `adapterWillPresentInterstitial`, that callback gets forwarded to publishers in the form of [adDidPresentFullScreenContent:](https://developers.google.com/admob/ios/api/reference/Protocols/GADFullScreenContentDelegate#-addidpresentfullscreencontent:). Publishers may depend on this callback to pause their apps. So if your adapter waits to call it and `vungleAdViewedForPlacement` comes 2 seconds later, that could mean the publisher's app pause mechanism could be delayed.

While it's nice to try and line up Google's mediation impressions for Vungle with the impressions counted in Vungle's dashbooard, the reality is that there will probably always be some discrepancy. I think the publisher getting addidpresentfullscreencontent: at the right time is actually more important for app flow.

I recommend mapping `vungleWillShowAdForPlacementID` to `adapterWillPresentInterstitial` right now. For Rewarded ads, which implement our newer mediation API, I recommend mapping `vungleWillShowAdForPlacementID` to [willPresentFullScreenView](https://developers.google.com/admob/ios/api/reference/Protocols/GADMediationAdEventDelegate.html#-willpresentfullscreenview) and `vungleAdViewedForPlacement` to [reportImpression](https://developers.google.com/admob/ios/api/reference/Protocols/GADMediationAdEventDelegate.html#-reportimpression). This enables the callbacks for rewarded to map 1:1 to what they're achieving. While today I believe we do still actually ping a rewarded mediation impression on either `willpresentfullscreenview` or `reportImpression`, whichever comes first, the mapping I recommend sets us up in the future for if we ever start to rely solely on reportImpression. And at some point in the future, we'd like to update the Vungle adapter to leverage the new mediation API for interstitials, at which point we can map `vungleAdViewedForPlacement` to `reportImpression` for interstitials too.

crbinsf:
@ericleich thanks so much for the reply - I'll follow up with product on our side and determine a path forward.

ericleich:
Hey @crbinsf , do you have any updates here?

stanley-vungle:
@googlebot I consent.

stanley-vungle:
Hey @ericleich , I've updated the PR with your previous suggestion by moving `adapterWillPresentInterstitial` into `willShowAd` rather than `didViewAd`. Thanks.

PR #297, opened 2021-03-03
Adding RewardedInterstitial Ad support for IronSource iOS Adapter.

chlandacastro:
Adding RewardedInterstitial Ad support for IronSource iOS Adapter.


chlandacastro:
@googlebot I signed it! and we'll verify it.

PR #305, opened 2021-03-31
[UnityAds] Adapter updates for 3.7.1 release

richardh-unity3d:


adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -102,91 +100,73 @@ - (void)presentFromViewController:(nonnull UIViewController *)viewController {
...
-    GADMAdapterUnityMapTableRemoveObjectForKey(_placementInUseRewarded, self->_placementID);
-  });
+- (void)unityAdsShowComplete:(NSString *)placementId withFinishState:(UnityAdsShowCompletionState)state {
+  [_adEventDelegate didEndVideo];
 
-  if (_adLoadCompletionHandler) {
-    NSError *error = GADUnityErrorWithDescription([NSString
-        stringWithFormat:@"Failed to load rewarded ad with placement ID '%@'", placementId]);
-    _adEventDelegate = _adLoadCompletionHandler(nil, error);
+  if (state == kUnityAdsFinishStateCompleted) {
```
chlandacastro, line 151: Rewarded is been skipped after completing videos.  
State is returning kUnityShowComplationStateCompleted 
Screen Shot 2021-04-01 at 1 28 26 PM
Screen Shot 2021-04-01 at 1 29 23 PM

richardh-unity3d: Thanks @chlandacastro, I'll look into this.

richardh-unity3d: Thanks again @chlandacastro, should be fixed now.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -102,91 +100,73 @@ - (void)presentFromViewController:(nonnull UIViewController *)viewController {
...
-  }
   [_adEventDelegate willPresentFullScreenView];
-  [UnityAds addDelegate:self];
-  [UnityAds show:viewController placementId:_placementID];
+  [UnityAds show:viewController placementId:_placementID showDelegate:self];
 }
 
-#pragma mark - UnityAdsExtendedDelegate Methods
+- (void)cleanup {
+  // Clean up references to prevent a memory leak where the object can not be cleand up
```
ericleich, line 108: Which object is leaking? who is holding onto the reference? This class's NSMapTable has a weak reference to GADMAdapterUnityRewardedAd. Does the UnityAds SDK keep a strong reference to the delegate passed into [UnityAds load]? If not, then there shouldn't be any leaks.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -102,91 +100,73 @@ - (void)presentFromViewController:(nonnull UIViewController *)viewController {
...
+  // due to holding on to the references.
+  _adEventDelegate = nil;
+  _adLoadCompletionHandler = nil;
```
ericleich, line 111: I prefer not to set completion handlers to nil due to possible crash if this is ever called without a nil check. If https://github.com/googleads/googleads-mobile-ios-mediation/blob/master/CONTRIBUTING.md#best-practices #7 is implemented on this class, the original completion handler is nilled out once we invoke the handler the first time, and there should be no need to set completion handler to nil.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -118,73 +118,28 @@ - (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewCo
     return;
   }
 
-  if (![UnityAds isReady:_placementID]) {
-    NSError *error = GADMAdapterUnityErrorWithCodeAndDescription(
-        GADMAdapterUnityErrorShowAdNotReady, @"Failed to show Unity Ads interstitial.");
-    [strongConnector adapter:strongAdapter didFailAd:error];
-    return;
+  if (!_placementID) {
```
ericleich, line 121: This is not invoking any failure callback. I'm assuming that [UnityAds show:rootViewController placementId:nil] is expected to fail and UnityAds SDK will invoke the unityAdsShowFailed callback?

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -118,73 +118,28 @@ - (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewCo
...
-- (void)unityAdsPlacementStateChanged:(NSString *)placementId
-                             oldState:(UnityAdsPlacementState)oldState
-                             newState:(UnityAdsPlacementState)newState {
-}
+#pragma mark - UnityAdsLoadDelegate Methods
 
-- (void)unityAdsDidFinish:(NSString *)placementID withFinishState:(UnityAdsFinishState)state {
-  [UnityAds removeDelegate:self];
+- (void)unityAdsAdLoaded:(nonnull NSString *)placementId {
+  _placementID = placementId;
```
ericleich, line 132: rename _placementID to _loadedPlacementID, as that is now how it's being used in this class.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -118,73 +118,28 @@ - (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewCo
...
-- (void)unityAdsReady:(nonnull NSString *)placementId {
-  // Logic to mark a placement ready has moved to the UnityAdsLoadDelegate function
-  // unityAdsAdLoaded.
-}
-
-#pragma mark - UnityAdsLoadDelegate Methods
-
-- (void)unityAdsAdFailedToLoad:(nonnull NSString *)placementId {
+- (void)unityAdsAdFailedToLoad:(NSString *)placementId withError:(UnityAdsLoadError)error withMessage:(NSString *)message {
+  _placementID = placementId;
```
ericleich, line 142: per my other comment of renaming the ivar to _loadedPlacementID, you can now set _loadedPlacementID to nil when the ad fails to load, and just use placementId directly through the rest of this method.

ericleich (changes requested):

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
richardh-unity3d: Thanks for the link @ericleich.  The suggested change fix the leak.  I will push an update shortly.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
richardh-unity3d: This is correct.  If the placement ID is nil, `unityAdsShowFailed` will be called with the error  kUnityShowErrorInvalidArgument.

Allowing show to be called here will help provide more insight into issues related to placement fill.  If a placement had been loaded but is no longer ready `unityAdsShowFailed` will be called with the error `kUnityShowErrorNotReady`.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
richardh-unity3d: I think we would prefer to still set the placement ID here so that if show is called despite the load failure, we will gain insight that this happened.  Otherwise show will be called with nil and we won't know what placement ID was attempting to be shown.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
ericleich: acknowledged. passing to UnityAds SDK also lets you get insight into how often this might be happening.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
ericleich: Ok, perhaps my advice to use "_loadedPlacementID" set from callbacks isn't good then.

Instead of trying to set a _placementID ivar from your callbacks, I suggest you just set the _placementID ivar to ```[[[strongConnector credentials] objectForKey:kGADMAdapterUnityPlacementID] copy];``` for interstitals, and ```_adConfiguration.credentials.settings[kGADMAdapterUnityPlacementID];``` for rewarded at the time when the ad is loaded. That ID comes from the string the user pasted into the AdMob UI for your "placementID" and is already the ID being used to load the ad. That way, you'll never get asked to show an ad with a nil placement, and if the ad isn't loaded at show time, your SDK is probably already logging something useful in the show failure callback.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -118,18 +118,18 @@ - (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewCo
     return;
   }
 
-  if (!_placementID) {
+  if (!_loadedPlacementID) {
     NSLog(@"Unity Ads received call to show before successfully loading an ad");
   }
   
   [strongConnector adapterWillPresentInterstitial:strongAdapter];
```
ericleich, line 125: Since your SDK is now responsible for sending "showFailed" callbacks, please move the equivalent "willPresent" callbacks into the unityAdsDidStart event. Please apply for interstitial and rewarded.

If the adapter sends back a "willPresent" callback before a "didFailToPresent" callback, the "didFailToPresent" never makes it back to the app, as "willPresent" is assumed to mean a successful present.

adapters/Unity/UnityAdapter/GADMUnityInterstitialAd.m
```diff
@@ -118,95 +118,91 @@ - (void)presentInterstitialFromRootViewController:(UIViewController *)rootViewCo
...
-    [strongNetworkConnector adapterDidReceiveInterstitial:strongAdapter];
+  if (!strongAdapter) {
+    return;
   }
+
+  NSError *errorWithDescription =
+      GADMAdapterUnitySDKErrorWithUnityAdsShowErrorAndMessage(error, message);
+  [strongConnector adapter:strongAdapter didFailAd:errorWithDescription];
+
+  [strongConnector adapterWillDismissInterstitial:strongAdapter];
```
ericleich, line 204: Per my other comment about moving adapterWillDismissInterstitial: to unityAdsShowStart:, this method should call adapterWillDismissInterstitial: just before adapterWillDismissInterstitial:, to simulate a failure since this interstitial mediation API doesn't have a true "show failure" event.

richardh-unity3d: I think you have a typo here, can you clarify?
> this method should call adapterWillDismissInterstitial: just before adapterWillDismissInterstitial

ericleich: Ah, I meant adapterWillPresentInterstitial: just before adapterWillDismissInterstitial:. Since this mediation API doesn't have to indicate "ad failed to show" we simulate that by a quick succession of presented/dismissed.

While you're here, you can also remove the `[strongConnector adapter:strongAdapter didFailAd:errorWithDescription];` call. `didFailAd:` is the load time error, so calling it from Unity's show-time callback will be too late.

adapters/Unity/UnityAdapter/GADMAdapterUnityRewardedAd.m
```diff
@@ -31,10 +31,11 @@ @implementation GADMAdapterUnityRewardedAd {
   GADMediationAdConfiguration *_adConfiguration;
 
   /// An ad event delegate to invoke when ad rendering events occur.
-  id<GADMediationRewardedAdEventDelegate> _adEventDelegate;
+  __weak id<GADMediationRewardedAdEventDelegate> _adEventDelegate;
```
ericleich, line 34: While delegate's are typically weak, we typically make this one strong in other adapters ([example](https://github.com/googleads/googleads-mobile-ios-mediation/blob/de680e4d4485d50949e5914bf6eefe91d2dbedff/adapters/Facebook/FacebookAdapter/OpenBidding/GADFBRewardedRenderer.m#L36-L39)). Unlike most delegate patterns where you pass a delegate to some other class, who should hold a weak reference to it, in this case Google's SDK is *returning* a delegate object to you from a completion handler. __weak only works because Google's SDK also happens to have a strong reference to the object it returned, but as a best practice we keep this explicitly strong just in case that ever changes.

Hopefully that doesn't affect any memory leak analysis you were doing, but please let me know if you noticed otherwise.

richardh-unity3d: We added `__weak` because without it we were seeing objects hanging around.  But I will take a look at the example you provided and double check what is going on without `__weak` just to be sure it is needed.

richardh-unity3d: @ericleich I just confirmed that if we don't do this then we have a retain cycle.

Screen Shot 2021-04-16 at 11 05 58 AM

We attempted to modify our code to be more similar to the example provided, but we still experienced the same retain cycle.

We then attempted to add a dispatch after to test breaking the retain cycle manually but it resulted a different retain cycle.

Screen Shot 2021-04-16 at 1 46 16 PM

```
- (void)loadRewardedAdForAdConfiguration:(GADMediationRewardedAdConfiguration *)adConfiguration
                       completionHandler:
                           (GADMediationRewardedLoadCompletionHandler)completionHandler {
  self.rewardedAd = [[GADMAdapterUnityRewardedAd alloc] init];
  [self.rewardedAd loadRewaredeAdWithAdConfiguration:adConfiguration completionHandler:completionHandler];
  
  NSInteger ttl = 10;
  dispatch_after(dispatch_time(DISPATCH_TIME_NOW, ttl * NSEC_PER_SEC), dispatch_get_main_queue(), ^{
    self.rewardedAd = nil;
  });
}
```

Do you have any other suggestions to try, or should we leave it `__weak` for now?

ericleich: @richardh-unity3d Thanks for the detailed writeup. `__weak` sounds good for now to unblock this PR, and `__weak` in general is still expected to work. I'll also sync with the team internally on this a bit more, as my previous understanding is strong would not cause this.

One interesting thing is I can't find even find GADMRewardedAdNetworkConnector in our SDK head code -- which Google SDK version are you using in your test?

richardh-unity3d: Thanks @ericleich.  It looks like my Podfile has `pod 'Google-Mobile-Ads-SDK', '~> 7.69.0'` in it.

ericleich: @richardh-unity3d got it. I confirmed `GADMRewardedAdNetworkConnector` was removed in version 8.0.0, but I'm still trying to unpack why it was referenced even in your test against 7.69.0. That class was used for our legacy rewarded medation API which the adapter doesn't implement.

Could you try running against 8.0.0 or higher and seeing if you still see the retain cycle with the strong reference? If you need a test app updated for version 8.0.0, you can use https://github.com/googleads/googleads-mobile-ios-examples/tree/master/Objective-C/admob/RewardedVideoExample.

ericleich: confiirmed internally that pre v8.0.0 there was some complexity in routing rewarded mediation, but in v8.0.0, that retain cycle shouldn't be there.

PR #306, opened 2021-04-03
Cleanup from AdColony iOS Open Bidding Banner PR

ericleich:
Cleanup from AdColony iOS Open Bidding Banner PR


brittmullen:
Looks good to us!

PR #308, opened 2021-04-08
Updating myTarget iOS Adapter (use MediaAdView in Native ad)

aserrr:


adapters/MyTarget/MyTargetAdapter/GADMAdapterMyTargetMediatedUnifiedNativeAd.m
```diff
@@ -17,6 +17,15 @@
 #import "GADMAdapterMyTargetExtraAssets.h"
 #import "GADMAdapterMyTargetUtils.h"
 
+@interface MTRGNativeAd ()
+
+- (void)registerView:(nonnull UIView *)containerView
```
ericleich, line 22: Is this API ever going to be in MyTarget's public headers? Redefining this in the adapter is fragile and could break in a future myTarget update?

aserrr: Well, this is a private API method we made specially for mediation adapter. Don't you mind if we add respondsToSelector() check?

ericleich: I see. Yeah, responds to selector would be better as it's more defensive.

aserrr: Here we go

ericleich: Change below looks good. Since you use respondsToSelector() now, can you remove the @interface MTRGNativeAd header declaration?

aserrr: I'd prefer to leave it in place in order to avoid 'undeclared selector' warning.

PR #316, opened 2021-05-04
Updating Fyber adapter to version 7.8.5.0

AviGelkop:


AviGelkop:
One important note: Fyber SDK will be built only with code 12.5+ (due to the use of SKAdNetwork 2.2)

adapters/Fyber/CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 ## Fyber iOS Mediation Adapter Changelog
 
+#### [Version 7.8.5.0]
```
ericleich, line 3: Can you update this to:

`#### Version 7.8.5.0 (in progress)`

The release doesn't actually go live until we do some further testing and create the payload, and the docs at https://developers.google.com/admob/ios/mediation/fyber#fyber-ios-mediation-adapter-changelog pull from this file. Users sometimes get confused because the changelog says a release is live before it actually is, so we've decided to use this "in progress" convention and we'll remove that text and add the payload link once the release is live.

ericleich:
@AviGelkop please take a look

adapters/Fyber/CHANGELOG.md
AviGelkop: OK no problem

PR #319, opened 2021-06-21
Chartboost SDK 8.4.2

daniel-barros:
Updated Chartboost adapter version string.
Just a reminder to get an adapters update released with support for the latest Chartboost SDK.


chlandacastro:
Thanks for sharing it, but we are working on this version internally already.

PR #320, opened 2021-07-21
[DRAFT/DO_NOT_MERGE][Vungle] Open Bidding Testing

stanley-vungle:
Adding the Open Bidding implementation to the AdMob adapter. GADMediationAdapterVungle now implements GADRTBAdapter instead of GADMediationAdapter. The VungleRouter implements the new VungleSDKHBDelegate. The Banner and Interstitial classes now implement their respective Mediation classes. The old interstitial class needs to remain as a router to the real adapter class, similar to the old rewarded class. The ad format classes are moved under "OpenBidding", but they can be moved under "AdFormats" instead. Arbitrary name.

### DIFF
--- a/adapters/Vungle/VungleAdapter/GADMediationAdapterVungle.m
+++ b/adapters/Vungle/VungleAdapter/GADMediationAdapterVungle.m
@@ -14,14 +14,22 @@
 
 #import "GADMediationAdapterVungle.h"
 #import "GADMAdapterVungleConstants.h"
-#import "GADMAdapterVungleRewardedAd.h"
+#import "GADMediationVungleBanner.h"
+#import "GADMediationVungleInterstitial.h"
+#import "GADMediationVungleRewardedAd.h"
 #import "GADMAdapterVungleRouter.h"
 #import "GADMAdapterVungleUtils.h"
 #import "VungleAdNetworkExtras.h"
 
 @implementation GADMediationAdapterVungle {
   /// Vungle rewarded ad wrapper.
-  GADMAdapterVungleRewardedAd *_rewardedAd;
+  GADMediationVungleRewardedAd *_rewardedAd;
+    
+  /// Vungle interstitial ad wrapper.
+  GADMediationVungleInterstitial *_interstitialAd;
+    
+  /// Vungle banner ad wrapper.
+  GADMediationVungleBanner *_bannerAd;
 }
 
 + (void)setUpWithConfiguration:(nonnull GADMediationServerConfiguration *)configuration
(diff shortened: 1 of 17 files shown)


adapters/Vungle/VungleAdapter/GADMediationAdapterVungle.m
```diff
@@ -14,14 +14,20 @@
...
+#import "GADMediationVungleRewardedAd.h"
 #import "GADMAdapterVungleRouter.h"
 #import "GADMAdapterVungleUtils.h"
 #import "VungleAdNetworkExtras.h"
 
 @implementation GADMediationAdapterVungle {
   /// Vungle rewarded ad wrapper.
-  GADMAdapterVungleRewardedAd *_rewardedAd;
+  GADMediationVungleRewardedAd *_rewardedAd;
+
```
joshh-devrel, line 27: Nit: missing `/// Vungle interstitial ad wrapper.`

adapters/Vungle/VungleAdapter/GADMediationAdapterVungle.m
```diff
@@ -14,14 +14,20 @@
...
 #import "GADMAdapterVungleUtils.h"
 #import "VungleAdNetworkExtras.h"
 
 @implementation GADMediationAdapterVungle {
   /// Vungle rewarded ad wrapper.
-  GADMAdapterVungleRewardedAd *_rewardedAd;
+  GADMediationVungleRewardedAd *_rewardedAd;
+    
+  GADMediationVungleInterstitial *_interstitialAd;
+
```
joshh-devrel, line 29: Nit: missing `/// Vungle banner ad wrapper.`

adapters/Vungle/VungleAdapter/OpenBidding/GADMediationVungleInterstitial.m
```diff
@@ -0,0 +1,193 @@
...
+    return;
+  }
+
+  NSString *appID = [GADMAdapterVungleUtils findAppID:_adConfiguration.credentials.settings];
+  if (!appID) {
+    NSError *error = GADMAdapterVungleErrorWithCodeAndDescription(
+        GADMAdapterVungleErrorInvalidServerParameters, @"Vungle app ID not specified.");
+    _adLoadCompletionHandler(nil, error);
+    return;
+  }
```
joshh-devrel, line 97: Question: Shouldn't we check the app ID before we do the `loadAd` above?

In addition, it seems like these few lines are also implemented in `GADMediationVungleBanner.m` and `GADMediationVungleRewardedAd.m`, perhaps we can have a standalone method for in `GADMediationAdapterVungle.m` and implement it in `loadInterstitialForAdConfiguration`, e.t.c.?

stanley-vungle: > Shouldn't we check the app ID before we do the loadAd above?
The app ID does not need to be checked in the loadAd above because the Vungle SDK has already initialized successfully. The app ID is only necessary for initialization and this section is to initialize the Vungle SDK if it hasn't been initialized yet.

> In addition, it seems like these few lines are also implemented in GADMediationVungleBanner.m and GADMediationVungleRewardedAd.m, perhaps we can have a standalone method in GADMediationAdapterVungle.m and implement it in loadInterstitialForAdConfiguration, e.t.c.?
These few lines are implemented in the banner and rewarded classes as well, but the `_adLoadCompletionHandler` is a different type. Perhaps a lazy initialization method should be added to the VungleRouter and have the delegate to handle failure passed in.

This case would be hit if the publisher does not call `[[GADMobileAds sharedInstance] startWithCompletionHandler:` and attempts to load an ad.

adapters/Vungle/VungleAdapter/OpenBidding/GADMediationVungleInterstitial.m
```diff
@@ -0,0 +1,187 @@
...
+        GADMAdapterVungleErrorAdAlreadyLoaded,
+        @"Only a maximum of one ad per placement can be requested from Vungle.");
+    _adLoadCompletionHandler(nil, error);
+    return;
+  }
+
+  if (![[GADMAdapterVungleRouter sharedInstance] isSDKInitialized]) {
+    NSString *appID = [GADMAdapterVungleUtils findAppID:_adConfiguration.credentials.settings];
+    [[GADMAdapterVungleRouter sharedInstance] initWithAppId:appID delegate:self];
+    return;
```
joshh-devrel, line 89: nit: Should we log a message that the interstitial ad isn't loaded due to SDK not initialized?

adapters/Vungle/VungleAdapter/OpenBidding/GADMediationVungleInterstitial.m
```diff
@@ -0,0 +1,187 @@
...
+
+- (void)loadAd {
+  NSError *error = [[GADMAdapterVungleRouter sharedInstance] loadAd:self.desiredPlacement
+                                                       withDelegate:self];
+  if (error) {
+    _adLoadCompletionHandler(nil, error);
+  }
+}
+
+#pragma mark - VungleRouter delegates
```
joshh-devrel, line 120: nit: should be `GADMAdapterVungleDelegate delegates`?

adapters/Vungle/VungleAdapter/OpenBidding/GADMediationVungleInterstitial.m
stanley-vungle: The SDK initialization here will queue the interstitial ad to be loaded and the Router class will call line 120, which would load the ad on success or return an error if the initialization failed. So I don't think a message is necessary here.
`(void)initialized:(BOOL)isSuccess error:(nullable NSError *)error`

PR #325, opened 2021-08-16
Update to Verizon SE v1.14.0

thomas-poland:


chlandacastro:
I finished testing this PR.  Only Native ads are not rendering.  I checked the previous version, which was working as expected, also not rendering Native ads.   It might be a server issue.

thomas-poland:
@chlandacastro RE: Native ads, I am seeing fills on the test ID: ca-app-pub-7866899558273871/9709724822 , also, we added updates for testing against v8.9.0

adapters/VerizonMedia/VerizonMediaAdapter/GADMAdapterVerizonBaseClass.m
```diff
@@ -345,7 +341,9 @@ - (void)setUserSettingsFromConnector {
 }
 
 - (void)setCoppaFromConnector {
-  VASAds.sharedInstance.COPPA = [_connector childDirectedTreatment];
+  VASDataPrivacyBuilder *builder = [[VASDataPrivacyBuilder alloc] initWithDataPrivacy:VASAds.sharedInstance.dataPrivacy];
+  builder.coppa.applies =  [[_connector childDirectedTreatment] boolValue];
+  VASAds.sharedInstance.dataPrivacy = [builder build];
```
ericleich, line 346: Users may supply privacy data directly to the adapter (e.g. https://developers.google.com/admob/ios/mediation/verizon-media#eu_consent_and_gdpr). Those docs will of course need to be updated now the the APIs are changing. However, this line looks like it will override the dataPrivacy passed in by a user. I suspect all of these instances I think will need to account for that. Possible solutions are: 

1) the adapter maintains a builder with all the info inside GADMVerizonPrivacy, that can keep being rebuilt any time this should be updated
2) Your API has some kind of [VASDataPrivacy toBuilder] method to get a builder with the current state

thomas-poland: @ericleich RE: docs - we’re reviewing our docs update and it will be released shortly, basically the docs outline what is done here - to rebuild privacy without changing other existing user settings, pass the current state to VASDataPrivacyBuilder initializer:

```VASDataPrivacyBuilder *builder = [[VASDataPrivacyBuilder alloc] initWithDataPrivacy:VASAds.sharedInstance.dataPrivacy];```

ericleich: @thomas-poland thanks for clarifying. Upon re-review, the code indeed is already implementing my suggestion #2.

Issue #329, opened 2021-09-05
Please include arm64-simulator slices in .xcframework


sgabello:
To fix warnings and errors in Xcode on M1 Macs.

chlandacastro:
The Following are the updated adapters that include arm64-simulator slice. 
AppLovin 4.7.2.1 ,Chartboost 8.5.0.2, Facebook 6.9.0.1, i-Mobile 2.2.0.1,  MoPub 5.18.2.1,

chlandacastro:
As of Wednesday, March 4, 2022. The Followings are the updated adapters that include the arm64-simulator slice:
AdColony 4.7.2.2
AppLovin 11.1.2.0
Chartboost 8.5.0.2,
Facebook 6.9.0.1,
i-Mobile 2.2.0.1,
inMobi 10.0.2.1
nend 7.2.0.0
Maio 1.6.0.1
MoPub 5.18.2.1.
MyTarget 5.15.0.1
Tapjoy 12.0.0.1
Unity 4.0.0.0
VerizonMedia 1.14.1.1

PR #330, opened 2021-09-05
Update adapters build script to include arm64-simulator slice

sgabello:
This will make the .xcframework work correctly on Apple Silicon M1 Macs.

Please note that YANDEX, VPON and MOBFOX build scripts don't use .xcframework and therefore I haven't implemented the changes for these adapters #329


sgabello:
Hello, could you review this pull request so that the adapters generated will include the slice for apple silicon MacBooks?

Thanks

sgabello:
@ericleich thanks you so much for the reply.

Just a couple of notes:

1. back in September when I first submit the PR there were not so many conflicts lol.
2. fair point but it shouldn't be your responsibility to make sure that arm64 is supported by other ad networks SDKs. If a network doesn't provide a slice, the app build process will fail on them and will push publishers to reach out to them to update their SDKs, instead of complaining with you ;)
3. I don't use CocoaPods - _I know... I might be one of the last few_ - so can't comment on that. But I thought CocoaPods would build the frameworks from source hence devs on Apple Silicon machines would not be impacted by this issue. If they are, I'm very surprised I'm the only one raising the problem.

In general I'm just really surprised no one is having this issue considering that M1 Macs have been around for more than one year now and AdMob is of course very popular among app developers.

To be honest it is not a big deal for me as I now have my fork that I use every time I need to update the SDKs... it's just a little time consuming to manually build the adapters every time. 

Mainly I was just trying to make life easier to other fellow devs that might get stuck on this and don't know how to do it. Maybe it would be a good idea to add a note on the official documentation to explain that arm64 support is being added and adapters can be built manually from source in the meanwhile. It took me quite a bit to figure it out. :)

Anyway... Thank you so much for taking the time to reply to my PR and explain why it can't be integrated at the moment. I really appreciated it! Thanks again!

PR #333, opened 2021-09-23
Vungle Bidding

stanley-vungle:
Adding the Open Bidding implementation to the AdMob adapter. GADMediationAdapterVungle now implements GADRTBAdapter instead of GADMediationAdapter. The VungleRouter implements the new VungleSDKHBDelegate. The Banner and Interstitial classes now implement their respective Mediation classes. The old interstitial class needs to remain as a router to the real adapter class, similar to the old rewarded class. The new ad format classes are moved under "Bidding". The legacy classes remain for waterfall cases. (#33)

Adding the Open Bidding implementation to the AdMob adapter. GADMediationAdapterVungle now implements GADRTBAdapter instead of GADMediationAdapter. The VungleRouter implements the new VungleSDKHBDelegate. The Banner and Interstitial classes now implement their respective Mediation classes. The old interstitial class needs to remain as a router to the real adapter class, similar to the old rewarded class. The new ad format classes are moved under "Bidding". The legacy classes remain for waterfall cases.

### DIFF
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
@@ -146,6 +146,11 @@
   return _bannerSize;
 }
 
+- (nullable NSString *)bidResponse {
+    // This is the waterfall interstitial section. It won't have a bid response
+    return nil;
+}
+
 - (void)initialized:(BOOL)isSuccess error:(nullable NSError *)error {
   if (!isSuccess) {
     [_connector adapter:_adapter didFailAd:error];
@@ -221,5 +226,6 @@
 - (void)rewardUser {
   // Do nothing.
 }
+
 
 @end
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
@@ -217,6 +217,7 @@
 
 - (void)trackClick {
   [_connector adapterDidGetAdClick:_adapter];
+  [_connector adapterWillPresentFullScreenModal:_adapter];
 }
 
 - (void)willLeaveApplication {
--- a/adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
+++ b/adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
@@ -147,7 +147,7 @@
 }
 
 - (nullable NSString *)bidResponse {
-    // This is the waterfall interstitial section. It won't have a bid response
+    // This is the waterfall banner section. It won't have a bid response.
     return nil;
 }
 
@@ -226,6 +226,5 @@
 - (void)rewardUser {
   // Do nothing.
 }
-
 
 @end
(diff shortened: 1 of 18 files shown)


adapters/Vungle/VungleAdapter/GADMAdapterVungleRewardBasedVideoAd.m
```diff
@@ -1,4 +1,4 @@
-// Copyright 2019 Google LLC
+// Copyright 2021 Google LLC
```
joshh-devrel, line 1: Ditto, nit, it should be: 
// Copyright 2019-2021 Google LLC

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -195,4 +207,5 @@ - (void)rewardUser {
   // Do nothing.
 }
 
+
```
joshh-devrel, line 210: Nit: Can remove empty line.

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
```diff
@@ -222,4 +227,5 @@ - (void)rewardUser {
   // Do nothing.
 }
 
+
```
joshh-devrel, line 230: Nit: Empty line can be removed?

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -32,6 +33,12 @@ @implementation GADMAdapterVungleInterstitial {
   BOOL _isAdLoaded;
 }
 
+// Redirect to the main adapter class for bidding
+// but still implement GADMAdNetworkAdapter for waterfall
```
joshh-devrel, line 37: Nit: Missing period at the end.

adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m
```diff
@@ -32,6 +33,12 @@ @implementation GADMAdapterVungleInterstitial {
...
++ (nonnull Class<GADMediationAdapter>)mainAdapterClass {
```
joshh-devrel, line 38: Question: When will this be called? This is also in `GADMAdapterVungleRewardBasedVideoAd` but not `GADMAdapterVungleBanner`.

stanley-vungle: This would be called by the Google SDK. From what I understand, the Google SDK looks for this legacy class when the publisher is trying to load an interstitial or banner and checks if this method is implemented. If it is implemented, the Google SDK would call this to get the adapter class and call the appropriate method for the ad format.

This is implemented in the `GADMAdapterVungleRewardBasedVideoAd` and here but not `GADMAdapterVungleBanner` because the Google SDK uses this class to load a banner as part of the older API.

stanley-vungle:
I had to convert the PR back to draft because I'm waiting for status from one of Vungle's internal testers.

PR #361, opened 2022-05-26
Added support in Vungle iOS adapter for loading multiple ads for the same placement ID in bidding rewarded ads.

joshh-devrel:
Added support in Vungle iOS adapter for loading multiple ads for the same placement ID in bidding rewarded ads.

### DIFF
--- a/adapters/Vungle/CHANGELOG.md
+++ b/adapters/Vungle/CHANGELOG.md
@@ -1,5 +1,13 @@
 ## Vungle iOS Mediation Adapter Changelog
 
+#### Next Version
+- Added support for loading multiple ads for the same placement ID in bidding interstitial and rewarded ads.
+- Verified compatibility with Google Mobile Ads SDK version 9.5.0.
+
+Built and tested with:
+- Google Mobile Ads SDK version 9.5.0.
+- Vungle SDK version 6.11.0
+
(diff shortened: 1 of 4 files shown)


stanley-vungle:
The load call of the interstitial class goes into the router that has other checks for tracking if there is a delegate being loaded already. If it passes that, it attempts to add the delegate into the tracking map. And if it fails because there's another delegate loaded there, then it will fail the load. Those 2 checks need to be bypassed for bidding. I think that it can be removed completely if the checks are moved to the GADMAdapterVungleBanner class.

https://github.com/googleads/googleads-mobile-ios-mediation/blob/3a4703b45f2b467271c49b155a9b3276dacda21b/adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleInterstitial.m#L103
https://github.com/googleads/googleads-mobile-ios-mediation/blob/3a4703b45f2b467271c49b155a9b3276dacda21b/adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m#L295

The router class needs to be refactored as well. The load calls can't be moved from the Router to the ad class since the v6 Vungle SDK has a singleton as a callback. So to load multiple ads, the delegate tracker needs to hold a reference to the bidding delegate instances to relay the Vungle SDK callbacks to the correct one.

PR #382, opened 2022-10-12
fix  main thread was stuck

wangyuanyuanyy:
Fix main thread was stuck caused by `[NSData dataWithContentsOfURL:url]`

### DIFF
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -88,25 +88,34 @@
                  strongSelf->_nativeAd = nativeAd;
                  strongSelf->_nativeAd.delegate = strongSelf;
                  strongSelf->_nativeAd.rootViewController = adConfiguration.topViewController;
-
-                 if (strongSelf->_loadCompletionHandler) {
-                   id<GADMediationNativeAdEventDelegate> delegate =
-                       strongSelf->_loadCompletionHandler(strongSelf, nil);
-                   strongSelf->_delegate = delegate;
-                 }
+      
+                 [strongSelf loadRequireData];
                }];
 }
 
-#pragma mark - GADMediationNativeAd
-
-- (GADNativeAdImage *)icon {
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
-    }
-  }
-  return _icon;
+- (void)loadRequireData {
+    NSString *urlString = _nativeAd.data.icon.imageURL;
+    dispatch_async(dispatch_get_global_queue(0, 0), ^{
+        NSURL *url = [NSURL URLWithString:urlString];
+        NSData *data = [NSData dataWithContentsOfURL:url];
+        GADPangleRTBNativeRenderer *__weak weakSelf = self;
+        dispatch_async(dispatch_get_main_queue(), ^{
+            GADPangleRTBNativeRenderer *strongSelf = weakSelf;
+            if (!strongSelf) {
+               return;
+            }
+            GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
+            strongSelf->_icon = image;
+            if (strongSelf->_loadCompletionHandler) {
+              id<GADMediationNativeAdEventDelegate> delegate =
+                  strongSelf->_loadCompletionHandler(strongSelf, nil);
+              strongSelf->_delegate = delegate;
+            }
+        });
+    });
 }
+
+#pragma mark - GADMediationNativeAd
 
 - (UIView *)mediaView {
   return _relatedView.mediaView;
@@ -178,13 +187,6 @@
 
 - (BOOL)handlesUserImpressions {
   return YES;
-}
-
-- (GADNativeAdImage *)imageWithUrlString:(NSString *)urlString {
-  NSURL *url = [NSURL URLWithString:urlString];
-  NSData *data = [NSData dataWithContentsOfURL:url];
-  UIImage *image = [UIImage imageWithData:data];
-  return [[GADNativeAdImage alloc] initWithImage:image];
 }
 
 - (void)didRenderInView:(nonnull UIView *)view
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -95,21 +95,27 @@
 
 - (void)loadRequireData {
     NSString *urlString = _nativeAd.data.icon.imageURL;
-    dispatch_async(dispatch_get_global_queue(0, 0), ^{
-        NSURL *url = [NSURL URLWithString:urlString];
+    if (!urlString.length) {
+        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        return;
+    }
+    NSURL *url = [NSURL URLWithString:urlString];
+    if (!url) {
+        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        return;
+    }
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{
         NSData *data = [NSData dataWithContentsOfURL:url];
-        GADPangleRTBNativeRenderer *__weak weakSelf = self;
         dispatch_async(dispatch_get_main_queue(), ^{
-            GADPangleRTBNativeRenderer *strongSelf = weakSelf;
-            if (!strongSelf) {
-               return;
+            if (!self) {
+                return;
             }
             GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
-            strongSelf->_icon = image;
-            if (strongSelf->_loadCompletionHandler) {
+            self->_icon = image;
+            if (self->_loadCompletionHandler) {
               id<GADMediationNativeAdEventDelegate> delegate =
-                  strongSelf->_loadCompletionHandler(strongSelf, nil);
-              strongSelf->_delegate = delegate;
+                  self->_loadCompletionHandler(self, nil);
+              self->_delegate = delegate;
             }
         });
     });
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -126,40 +126,40 @@
   return _icon;
 }
 
-- (UIView *)mediaView {
+- (nullable UIView *)mediaView {
   return _relatedView.mediaView;
 }
 
-- (UIView *)adChoicesView {
+- (nullable UIView *)adChoicesView {
   return _relatedView.logoADImageView;
 }
 
-- (NSString *)headline {
+- (nullable NSString *)headline {
   if (_nativeAd && _nativeAd.data) {
     return _nativeAd.data.AdTitle;
   }
   return nil;
 }
 
-- (NSString *)body {
+- (nullable NSString *)body {
   if (_nativeAd && _nativeAd.data) {
     return _nativeAd.data.AdDescription;
   }
   return nil;
 }
 
-- (NSString *)callToAction {
+- (nullable NSString *)callToAction {
   if (_nativeAd && _nativeAd.data) {
     return _nativeAd.data.buttonText;
   }
   return nil;
 }
 
-- (NSDecimalNumber *)starRating {
+- (nullable NSDecimalNumber *)starRating {
   return nil;
 }
 
-- (NSArray<GADNativeAdImage *> *)images {
+- (nullable NSArray<GADNativeAdImage *> *)images {
   return nil;
 }
 
@@ -167,18 +167,18 @@
   return nil;
 }
 
-- (NSString *)price {
+- (nullable NSString *)price {
   return nil;
 }
 
-- (NSString *)advertiser {
+- (nullable NSString *)advertiser {
   if (_nativeAd && _nativeAd.data) {
     return _nativeAd.data.AdTitle;
   }
   return nil;
 }
 
-- (NSDictionary<NSString *, id> *)extraAssets {
+- (nullable NSDictionary<NSString *, id> *)extraAssets {
   return nil;
 }
 
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -94,14 +94,21 @@
 }
 
 - (void)loadRequiredData {
+    GADPangleRTBNativeRenderer *__weak weakSelf = self;
+    void (^localBlock)(void) = ^{
+        GADPangleRTBNativeRenderer *strongSelf = weakSelf;
+        if (strongSelf && strongSelf->_loadCompletionHandler) {
+            strongSelf->_delegate = strongSelf->_loadCompletionHandler(strongSelf,nil);
+        }
+    };
     NSString *URLString = _nativeAd.data.icon.imageURL;
     if (!URLString.length) {
-        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        localBlock();
         return;
     }
     NSURL *url = [NSURL URLWithString:URLString];
     if (!url) {
-        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        localBlock();
         return;
     }
     NSURLSession *session = [NSURLSession sharedSession];
@@ -112,11 +119,7 @@
             }
             GADNativeAdImage *image = (!error && data) ? [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]] : nil;
             self->_icon = image;
-            if (self->_loadCompletionHandler) {
-              id<GADMediationNativeAdEventDelegate> delegate =
-                  self->_loadCompletionHandler(self, nil);
-              self->_delegate = delegate;
-            }
+            localBlock();
         });
     }];
     [task resume];
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -104,13 +104,13 @@
         if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
         return;
     }
-    dispatch_async(dispatch_get_global_queue(0, 0), ^{
-        NSData *data = [NSData dataWithContentsOfURL:url];
+    NSURLSession *session = [NSURLSession sharedSession];
+    NSURLSessionDataTask *task = [session dataTaskWithURL:url completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
         dispatch_async(dispatch_get_main_queue(), ^{
             if (!self) {
                 return;
             }
-            GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
+            GADNativeAdImage *image = (!error && data) ? [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]] : nil;
             self->_icon = image;
             if (self->_loadCompletionHandler) {
               id<GADMediationNativeAdEventDelegate> delegate =
@@ -118,7 +118,8 @@
               self->_delegate = delegate;
             }
         });
-    });
+    }];
+    [task resume];
 }
 
 #pragma mark - GADMediationNativeAd
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -122,6 +122,9 @@
 }
 
 #pragma mark - GADMediationNativeAd
+- (nullable GADNativeAdImage *)icon {
+  return _icon;
+}
 
 - (UIView *)mediaView {
   return _relatedView.mediaView;
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -89,17 +89,17 @@
                  strongSelf->_nativeAd.delegate = strongSelf;
                  strongSelf->_nativeAd.rootViewController = adConfiguration.topViewController;
       
-                 [strongSelf loadRequireData];
+                 [strongSelf loadRequiredData];
                }];
 }
 
-- (void)loadRequireData {
-    NSString *urlString = _nativeAd.data.icon.imageURL;
-    if (!urlString.length) {
+- (void)loadRequiredData {
+    NSString *URLString = _nativeAd.data.icon.imageURL;
+    if (!URLString.length) {
         if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
         return;
     }
-    NSURL *url = [NSURL URLWithString:urlString];
+    NSURL *url = [NSURL URLWithString:URLString];
     if (!url) {
         if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
         return;
--- a/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
+++ b/adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
@@ -114,9 +114,6 @@
     NSURLSession *session = [NSURLSession sharedSession];
     NSURLSessionDataTask *task = [session dataTaskWithURL:url completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
         dispatch_async(dispatch_get_main_queue(), ^{
-            if (!self) {
-                return;
-            }
             GADNativeAdImage *image = (!error && data) ? [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]] : nil;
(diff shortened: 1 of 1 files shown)


adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,35 @@ - (void)renderNativeAdForAdConfiguration:
...
 
-#pragma mark - GADMediationNativeAd
-
-- (GADNativeAdImage *)icon {
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
-    }
-  }
-  return _icon;
```
joshh-devrel, line 108: Why is this removed?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,35 @@ - (void)renderNativeAdForAdConfiguration:
...
+- (void)loadRequireData {
+    NSString *urlString = _nativeAd.data.icon.imageURL;
```
joshh-devrel, line 97: Is it guaranteed that we can get the `imageURL` successfully every time through accessing the property?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,35 @@ - (void)renderNativeAdForAdConfiguration:
...
+    dispatch_async(dispatch_get_global_queue(0, 0), ^{
```
joshh-devrel, line 98: Seems like we are having the priority `DISPATCH_QUEUE_PRIORITY_DEFAULT` here? Can you explain why this is chosen against the other priorities (e.g. `DISPATCH_QUEUE_PRIORITY_HIGH`)?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,35 @@ - (void)renderNativeAdForAdConfiguration:
...
-#pragma mark - GADMediationNativeAd
-
-- (GADNativeAdImage *)icon {
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
-    }
-  }
-  return _icon;
+- (void)loadRequireData {
```
joshh-devrel, line 96: We probably can move all the implementations to above (i.e. line `92`) without having an extra method.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,35 @@ - (void)renderNativeAdForAdConfiguration:
...
+    NSString *urlString = _nativeAd.data.icon.imageURL;
+    dispatch_async(dispatch_get_global_queue(0, 0), ^{
+        NSURL *url = [NSURL URLWithString:urlString];
+        NSData *data = [NSData dataWithContentsOfURL:url];
+        GADPangleRTBNativeRenderer *__weak weakSelf = self;
+        dispatch_async(dispatch_get_main_queue(), ^{
+            GADPangleRTBNativeRenderer *strongSelf = weakSelf;
+            if (!strongSelf) {
+               return;
+            }
```
joshh-devrel, line 106: Do we really need the `weakSelf` and `strongSelf` implementations here? We already do that above.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
joshh-devrel: Do we need the check below?
```
if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil)
```

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -180,13 +189,6 @@ - (BOOL)handlesUserImpressions {
   return YES;
 }
 
-- (GADNativeAdImage *)imageWithUrlString:(NSString *)urlString {
-  NSURL *url = [NSURL URLWithString:urlString];
-  NSData *data = [NSData dataWithContentsOfURL:url];
-  UIImage *image = [UIImage imageWithData:data];
-  return [[GADNativeAdImage alloc] initWithImage:image];
-}
-
```
joshh-devrel, line 189: Why is this method removed? Seems like we can use it above?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,35 @@ - (void)renderNativeAdForAdConfiguration:
...
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
-    }
-  }
-  return _icon;
+- (void)loadRequireData {
+    NSString *urlString = _nativeAd.data.icon.imageURL;
+    dispatch_async(dispatch_get_global_queue(0, 0), ^{
+        NSURL *url = [NSURL URLWithString:urlString];
+        NSData *data = [NSData dataWithContentsOfURL:url];
```
joshh-devrel, line 100: You mentioned that the thread was stuck because of this line. Can you clarify about this? Should there be a fix in Pangle SDK if the url cannot be populated?

wangyuanyuanyy: `Don't use this synchronous method to request network-based URLs. For network-based URLs, this method can block the current thread for tens of seconds on a slow network, resulting in a poor user experience, and in iOS, may cause your app to be terminated.` in apple development doc

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: It moved in `- (void)loadRequireData`

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: I will remove it.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: Will this nest too many levels? Code is less readable.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: ok~

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: I changed the code.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: The `_icon` set in `-loadRequireData`

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
...
-                   strongSelf->_delegate = delegate;
-                 }
+      
+                 [strongSelf loadRequireData];
                }];
 }
 
-#pragma mark - GADMediationNativeAd
-
-- (GADNativeAdImage *)icon {
```
joshh-devrel, line 102: We need this method to map the `icon` asset.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
...
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
+- (void)loadRequireData {
```
joshh-devrel, line 96: I think we should still move this method to above.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
...
+        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        return;
     }
-  }
-  return _icon;
+    NSURL *url = [NSURL URLWithString:urlString];
+    if (!url) {
+        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        return;
+    }
```
joshh-devrel, line 106: We shouldn't need this check if `urlString` is valid?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
...
+    dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{
```
joshh-devrel, line 107: Can you explain why it is `HIGH` now?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
joshh-devrel: Seems like this is the documentation: https://developer.apple.com/documentation/foundation/nsdata/1547245-datawithcontentsofurl

Then shouldn't we use `[dataTaskWithURL:completionHandler:](https://developer.apple.com/documentation/foundation/nsurlsession/1410330-datataskwithurl)` instead?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
joshh-devrel: We can keep this method.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,26 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
...
-
-- (GADNativeAdImage *)icon {
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
+- (void)loadRequireData {
+    NSString *urlString = _nativeAd.data.icon.imageURL;
+    if (!urlString.length) {
+        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
+        return;
```
joshh-devrel, line 100: What to do with the `icon` in this case?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
joshh-devrel, line 102: Also, shouldn't all the changes / fixes be added here instead since the issue is related to the icon?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: This method  was used to load icon image, but now the loading progress is implemented like this:
> GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
> self->_icon = image;

so we don't need this method anymore

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: `- loadRequireData` is invoked when PAGLNativeAd loadAd callback success.So _nativeAd.data isn't nil.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: We use `-dataWithContentsOfURL` asynchronously. Must we use  `[dataTaskWithURL:completionHandler:]`?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: `_icon` set is implement like this, in method`- (void)loadRequireData`
> GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
>  self->_icon = image;

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: You mean moving  `-loadRequireData`  line 96 to  line 105?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: `[NSURL URLWithString:urlString]` is return `An NSURL object initialized with URLString. If the URL string was malformed or nil, returns nil.`

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: In this case icon is nil. In `GADMediationNativeAd` protocol, icon is a nullable property.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
joshh-devrel, line 102: As the previous comment was resolved, I am going to add another one.

Please keep this method as it's for mapping the icon asset, also see [the API documentation](https://developers.google.com/ad-manager/mobile-ads-sdk/ios/api/reference/Protocols/GADMediatedUnifiedNativeAd.html#icon).

We should do:
```
- (nullable GADNativeAdImage *)icon {
  return _icon;
}
```

wangyuanyuanyy: ok~

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,23 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
...
+              self->_delegate = delegate;
+            }
+        });
+    });
+}
+
+#pragma mark - GADMediationNativeAd
+- (nullable GADNativeAdImage *)icon {
   return _icon;
 }
```
joshh-devrel, line 128: For all the assets, can you add a `nullable` annotation just like what you did for `icon`?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -180,13 +198,6 @@ - (BOOL)handlesUserImpressions {
   return YES;
 }
 
-- (GADNativeAdImage *)imageWithUrlString:(NSString *)urlString {
-  NSURL *url = [NSURL URLWithString:urlString];
-  NSData *data = [NSData dataWithContentsOfURL:url];
-  UIImage *image = [UIImage imageWithData:data];
-  return [[GADNativeAdImage alloc] initWithImage:image];
-}
-
```
joshh-devrel, line 189: Can we do the below instead?
```
- (GADNativeAdImage *)loadIconWithUrlString:(NSString *)urlString {
  NSURL *url = [NSURL URLWithString:urlString];
  dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_HIGH, 0), ^{
    NSData *data = [NSData dataWithContentsOfURL:url];
    dispatch_async(dispatch_get_main_queue(), ^{
      GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
    });
    _icon = image;
  });
}

File: adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m [hunk 15]
```diff
@@ -88,23 +88,41 @@ - (void)renderNativeAdForAdConfiguration:
                  strongSelf->_nativeAd = nativeAd;
                  strongSelf->_nativeAd.delegate = strongSelf;
                  strongSelf->_nativeAd.rootViewController = adConfiguration.topViewController;
-
-                 if (strongSelf->_loadCompletionHandler) {
-                   id<GADMediationNativeAdEventDelegate> delegate =
-                       strongSelf->_loadCompletionHandler(strongSelf, nil);
-                   strongSelf->_delegate = delegate;
-                 }
```
joshh-devrel, line 96: I have added a new method `loadIconWithUrlString` in another comment.
Then in here can we revert this change and call that method instead?

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: `_icon = image` will report an error because the image is not within the scope of the current position. And this method will cause a delay in getting the icon.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: I replied to your comment.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: ok~

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: In the snippet, the `_icon = nil` obtained when invoke `_loadCompletionHandler`

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,79 +88,97 @@ - (void)renderNativeAdForAdConfiguration:
...
+            }
+            GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
+            self->_icon = image;
+            if (self->_loadCompletionHandler) {
+              id<GADMediationNativeAdEventDelegate> delegate =
+                  self->_loadCompletionHandler(self, nil);
+              self->_delegate = delegate;
+            }
+        });
+    });
```
joshh-devrel, line 121: @wangyuanyuanyy 

Since the URL is a network URL, can you update the implementation with [dataTaskWithURL:completionHandler:](https://developer.apple.com/documentation/foundation/nsurlsession/1410330-datataskwithurl) method of the [NSURLSession](https://developer.apple.com/library/archive/technotes/tn2232/_index.html#//apple_ref/doc/uid/DTS40012884-CH1-SECNSURLSESSION) class?

Even on concurrent queues, running long synchronous methods is not ideal. Since there's a limit of 64 concurrent processes across all concurrent queues and it can fill up suprisingly quick with long sync calls.

wangyuanyuanyy: Do you mean the code is written like this ?
> NSURLSession *session = [NSURLSession sharedSession];
    [session dataTaskWithURL:url completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
        dispatch_async(dispatch_get_main_queue(), ^{
            if (!self) {
                return;
            }
            GADNativeAdImage *image = [[GADNativeAdImage alloc] initWithImage:[UIImage imageWithData:data]];
            self->_icon = image;
            if (self->_loadCompletionHandler) {
              id<GADMediationNativeAdEventDelegate> delegate =
                  self->_loadCompletionHandler(self, nil);
              self->_delegate = delegate;
            }
        });
    }];

joshh-devrel: @wangyuanyuanyy I haven't looked into the detailed implementation yet, but once you make the change to use [dataTaskWithURL:completionHandler:](https://developer.apple.com/documentation/foundation/nsurlsession/1410330-datataskwithurl) method of the [NSURLSession](https://developer.apple.com/library/archive/technotes/tn2232/_index.html#//apple_ref/doc/uid/DTS40012884-CH1-SECNSURLSESSION) class, we can review the code again.

wangyuanyuanyy: Ok~ I changed .

wangyuanyuanyy:
Hello, I changed method of the load image data. Please review the again.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,79 +88,98 @@ - (void)renderNativeAdForAdConfiguration:
...
 }
 
-#pragma mark - GADMediationNativeAd
-
-- (GADNativeAdImage *)icon {
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
+- (void)loadRequireData {
+    NSString *urlString = _nativeAd.data.icon.imageURL;
```
joshh-devrel, line 97: `URLString`, acronyms in ObjC are always uppercase.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,79 +88,98 @@ - (void)renderNativeAdForAdConfiguration:
...
+    if (!urlString.length) {
+        if (_loadCompletionHandler) _delegate = _loadCompletionHandler(self,nil);
```
joshh-devrel, line 99: Rather than needing to wrap this in every location, we can create a block locally at the top like:

```
localBlock = ^{(NSError *error)if (_loadCompletionHandler) {
   _loadCompletionHandler(self,error);
}};
```
then everywhere below can call localBlock without a guard and not risk crashing.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,79 +88,98 @@ - (void)renderNativeAdForAdConfiguration:
...
                }];
 }
 
-#pragma mark - GADMediationNativeAd
-
-- (GADNativeAdImage *)icon {
-  if (!_icon) {
-    if (_nativeAd.data.icon && _nativeAd.data.icon.imageURL != nil) {
-      _icon = [self imageWithUrlString:_nativeAd.data.icon.imageURL];
+- (void)loadRequireData {
```
joshh-devrel, line 96: Nit, `loadRequiredData` or `loadNativeData`.

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: ok~

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: Ok~

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
wangyuanyuanyy: Ok~

adapters/Pangle/PangleAdapter/Bidding/GADPangleRTBNativeRenderer.m
```diff
@@ -88,79 +88,101 @@ - (void)renderNativeAdForAdConfiguration:
...
-  }
+    NSURL *url = [NSURL URLWithString:URLString];
+    if (!url) {
+        localBlock();
+        return;
+    }
+    NSURLSession *session = [NSURLSession sharedSession];
+    NSURLSessionDataTask *task = [session dataTaskWithURL:url completionHandler:^(NSData * _Nullable data, NSURLResponse * _Nullable response, NSError * _Nullable error) {
+        dispatch_async(dispatch_get_main_queue(), ^{
+            if (!self) {
```
joshh-devrel, line 117: I assume that this is intended to be another `weakSelf` / `strongSelf` construct. `self` will always be `nonnull` here because this is retaining it.

wangyuanyuanyy: ok~

PR #390, opened 2023-01-11
Add arm64 arch for IronSourceAdapter xcframework

ValeriyMaliuk:
Since IronSource 2.6.0 adds support of arm64 arch for simulator - add it for adapter


joshh-devrel:
We have added support for arm64 simulator architecture in ironSource iOS adapter [version 7.2.7.0.0](https://developers.google.com/admob/ios/mediation/ironsource#version-7.2.7.0.0). Thanks!

PR #395, opened 2023-01-30
Adapter Changes for Vungle SDK 7.0.0

stanley-vungle:
This PR is integrating Vungle (Liftoff) 7.0 SDK with the Admob adapter.


adapters/Vungle/VungleAdapter/VungleRouterConsent.m
```diff
@@ -13,23 +13,24 @@
...
+}
+
++ (void)updateCOPPAStatus:(BOOL)optIn {
+  [VunglePrivacySettings setCOPPAStatus:optIn];
 }
 
-+ (VungleConsentStatus)getConsentStatus {
-  return currentConsentStatus;
++ (void)updateIDFVStatus:(BOOL)optIn {
+  [VunglePrivacySettings setPublishIdfv:optIn];
```
joshuajplagonera, line 33: If these methods are static and can be called at any time (before or after initialization), I think publishers can just call them directly instead of needing a wrapper API?
If so, then we can just remove this class entirely.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -49,501 +37,55 @@ + (nonnull GADMAdapterVungleRouter *)sharedInstance {
...
-  _isInitializing = NO;
-  NSMapTable<NSString *, id<GADMAdapterVungleDelegate>> *delegates = [_initializingDelegates copy];
-  for (NSString *key in delegates) {
-    id<GADMAdapterVungleDelegate> delegate = [delegates objectForKey:key];
-    [delegate initialized:isSuccess error:error];
-  }
-
-  [_initializingDelegates removeAllObjects];
-  _prioritizedPlacementID = nil;
+- (BOOL)isSDKInitialized {
```
joshuajplagonera, line 83: We can probably remove this method and just call `[VungleAds isInitialized]` directly.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
```diff
@@ -49,501 +37,55 @@ + (nonnull GADMAdapterVungleRouter *)sharedInstance {
...
+  return [VungleAds isInitialized];
 }
 
 - (NSString *)getSuperToken {
```
joshuajplagonera, line 87: Similar to this, I think we can just directly call `[VungleAds getBiddingToken]` instead of needing this util method.

adapters/Vungle/VungleAdapter/GADMAdapterVungleUtils.h
```diff
@@ -13,44 +13,28 @@
...
+                                                               NSInteger vungleCode,
+                                                               NSString *_Nonnull description);
+
+/// Returns a NSError with code |code| and with NSLocalizedDescriptionKey and
+/// NSLocalizedFailureReasonErrorKey values set to |description|, specifically for invalid placement id.
+NSError *_Nonnull GADMAdapterVungleInvalidPlacementErrorWithCodeAndDescription(void);
+
+/// Returns a NSError with code |code| and with NSLocalizedDescriptionKey and
+/// NSLocalizedFailureReasonErrorKey values set to |description|, specifically for invalid app id.
+NSError *_Nonnull GADMAdapterVungleInvalidAppIdErrorWithCodeAndDescription(void);
```
joshuajplagonera, line 37: Instead of creating a separate utility method for every adapter error code, we can just create one utility method that takes in an adapter error code.

See https://github.com/googleads/googleads-mobile-ios-mediation/blob/main/adapters/Mintegral/MintegralAdapter/GADMAdapterMintegralUtils.m#L20 for reference.

joshuajplagonera (changes requested):
Initial review.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,133 @@ - (void)requestBannerAd {
...
+    [_bannerAd presentOn:_bannerView];
   }
+}
 
-  if (!_delegate) {
-    [[GADMAdapterVungleBiddingRouter sharedInstance] removeDelegate:self];
-    return;
-  }
+- (void)bannerAdDidFailToLoad:(VungleBanner *)banner withError:(NSError *)error {
+  NSError *gadError = GADMAdapterVungleErrorToGADError(GADMAdapterVungleErrorAdNotPlayable,
```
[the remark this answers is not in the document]
stanley-vungle: The errors from Vungle SDK has been updated to have more precise error codes. So it will have a lot more values than the ones listed on the doc and newer codes may be made in the future.

So if this is removed, then the [documentation here](https://developers.google.com/admob/ios/mediation/vungle#error_codes) would need to be removed or updated to the latest. I think having the adapter map it to an existing code would be simpler on both our ends.

adapters/Vungle/VungleAdapter/GADMAdapterVungleUtils.h
stanley-vungle: There is a single utility method that takes in an adapter error code.
https://github.com/googleads/googleads-mobile-ios-mediation/blob/f21398afb53030c06ec44785fdb12802b1140d2d/adapters/Vungle/VungleAdapter/GADMAdapterVungleUtils.m#L18

Those utility method are convenience methods because the code and message is the same and it is utilized at multiple locations. These convenience methods use the one utility method. For example.
https://github.com/googleads/googleads-mobile-ios-mediation/blob/f21398afb53030c06ec44785fdb12802b1140d2d/adapters/Vungle/VungleAdapter/GADMAdapterVungleInterstitial.m#L95
https://github.com/googleads/googleads-mobile-ios-mediation/blob/f21398afb53030c06ec44785fdb12802b1140d2d/adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m#L62

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.h
```diff
@@ -21,13 +21,10 @@
     initWithAdConfiguration:(nonnull GADMediationBannerAdConfiguration *)adConfiguration
           completionHandler:(nonnull GADMediationBannerLoadCompletionHandler)handler;
 
-/// Unavailable.
+/// Constructor is unavailable. Please use initWithAdConfiguration:completionHandler:.
 - (nonnull instancetype)init NS_UNAVAILABLE;
 
-/// Requests a banner ad from Vungle.
+/// Requests a interstitial ad from Vungle.
```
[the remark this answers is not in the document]
stanley-vungle: Updated.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -62,13 +60,12 @@ - (void)dealloc {
   self = [super init];
   if (self) {
     _adConfiguration = adConfiguration;
-    _bannerSize = GADMAdapterVungleAdSizeForAdSize([adConfiguration adSize]);
+    _bannerSize = [self vungleAdSizeForAdSize:[adConfiguration adSize]];
```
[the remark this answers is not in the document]
stanley-vungle: Updated.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
stanley-vungle: Updated.

adapters/Vungle/VungleAdapter/GADMAdapterVungleRouter.m
stanley-vungle: Updated.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,133 @@ - (void)requestBannerAd {
...
-  self.bannerState = BannerRouterDelegateStateWillPlay;
-  NSError *error = [self renderAd];
-  if (error) {
-    _adLoadCompletionHandler(nil, error);
-    return;
-  }
-
   if (_adLoadCompletionHandler) {
     _delegate = _adLoadCompletionHandler(self, nil);
+    [_bannerAd presentOn:_bannerView];
```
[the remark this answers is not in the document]
stanley-vungle: Updated.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,133 @@ - (void)requestBannerAd {
...
-                              adMarkup:[self bidResponse]
-                                 error:&bannerError];
-  return bannerError;
+  return GADAdSizeInvalid;
 }
 
-- (void)cleanUp {
-  if (_didBannerFinishPresenting) {
-    return;
+- (BannerSize)convertGADAdSizeToBannerSize {
```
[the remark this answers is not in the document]
stanley-vungle: Yes. It can.
The Vungle SDK doesn't have an "invalid" in the banner size enum so it would default to our regular size banner and then would fail presentation because of a size mismatch instead of failing the load request.

Before I change it, can I get a thought about whether it is better to fail on the load vs the playback?

adapters/Vungle/VungleAdapter/VungleRouterConsent.m
stanley-vungle: The methods in the Vungle SDK are static. We have a publisher request to be able to set the privacy values through the adapter code rather than communicating directly with the SDK API, which was why this class was created before to set the GDPR status. The RouterConfiguration class was removed to condense the IDFV setting to this class too.

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
```diff
@@ -24,25 +24,26 @@
...
+@property(nonatomic, copy) NSArray<NSString *> *_Nullable allPlacements DEPRECATED_ATTRIBUTE;
 
-@property(nonatomic, copy, readonly) NSString *_Nonnull UUID;
+@property(nonatomic, copy) NSNumber *_Nullable orientations DEPRECATED_ATTRIBUTE;
 
-@property (nonatomic, readonly, assign) BOOL muteIsSet;
+@property(nonatomic, copy, readonly) NSString *_Nonnull UUID DEPRECATED_ATTRIBUTE;
 
-@property (nonatomic, assign) NSInteger nativeAdOptionPosition;
+@property (nonatomic, readonly, assign) BOOL muteIsSet DEPRECATED_ATTRIBUTE;
```
joshh-devrel, line 47: Since all of these properties have been marked deprecated, can you confirm if they aren't supported anymore? Are there substitutions?

(If they aren't supported anymore, this section in the dev doc won't be valid anymore: https://developers.google.com/admob/ios/mediation/vungle#network-specific_parameters)

adapters/Vungle/Public/Headers/VungleRouterConsent.h
```diff
@@ -13,9 +13,16 @@
...
+
+/// Updates the GDPR status in the Vungle SDK
++ (void)updateGDPRStatus:(BOOL)optIn;
+
+/// Updates the CCPA status in the Vungle SDK
++ (void)updateCCPAStatus:(BOOL)optIn;
+
+/// Determines whether the IDFV can be collected or not
++ (void)updateIDFVStatus:(BOOL)optIn;
+
```
joshh-devrel, line 27: When should these consent values be passed to the Vungle SDK? Is it before Vungle SDK initializes or at ad request time?

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,130 @@ - (void)requestBannerAd {
...
-    return;
-  }
-  _adLoadCompletionHandler(nil, error);
+- (void)bannerAdWillPresent:(VungleBanner *)banner {
+  // No-op.
 }
 
-- (void)willShowAd {
-  self.bannerState = BannerRouterDelegateStatePlaying;
+- (void)bannerAdDidPresent:(VungleBanner *)banner {
```
joshh-devrel, line 185: Will this callback be called when banner ad is presented? Or when the modal is presented after the banner ad is clicked?

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,130 @@ - (void)requestBannerAd {
...
 }
 
-- (void)adNotAvailable:(nonnull NSError *)error {
-  if (_isAdLoaded) {
-    // Already invoked an ad load callback.
-    return;
-  }
-  _adLoadCompletionHandler(nil, error);
+- (void)bannerAdWillPresent:(VungleBanner *)banner {
+  // No-op.
```
joshh-devrel, line 182: Is this no-op or just there isn't an equivalent callback in Google Mobile Ads SDK?

If the case is latter, we should update it to:
```
    // Google Mobile Ads SDK doesn't have a matching event.
```

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleInterstitial.m
```diff
@@ -91,98 +93,74 @@ - (void)requestInterstitialAd {
...
 }
 
-- (void)willShowAd {
+- (void)interstitialAdWillPresent:(VungleInterstitial *)interstitial {
   [_delegate willPresentFullScreenView];
 }
 
-- (void)didShowAd {
-  // Do nothing.
+- (void)interstitialAdDidPresent:(VungleInterstitial *)interstitial {
```
joshh-devrel, line 125: Question, do this callback and `interstitialAdDidTrackImpression` call at the same time?

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
```diff
@@ -29,16 +29,14 @@ @implementation GADMAdapterVungleBanner {
   /// The requested ad size.
   GADAdSize _bannerSize;
 
-  /// Indicates whether the banner ad finished presenting.
-  BOOL _didBannerFinishPresenting;
+  /// Vungle Banner instance
```
joshh-devrel, line 32: Nit:
```
/// Vungle banner ad instance.
```

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
```diff
@@ -29,16 +29,14 @@ @implementation GADMAdapterVungleBanner {
...
+  VungleBanner *_bannerAd;
+
+  /// UIView for Vungle to mount the ad
```
joshh-devrel, line 35: Maybe can update to below:
```
/// UIView that contains Vungle banner ad.
```

adapters/Vungle/VungleAdapter/VungleRouterConsent.m
```diff
@@ -13,23 +13,24 @@
...
++ (void)updateGDPRStatus:(BOOL)optIn {
+  [VunglePrivacySettings setGDPRStatus:optIn];
+}
+
++ (void)updateCCPAStatus:(BOOL)optIn {
+  [VunglePrivacySettings setCCPAStatus:optIn];
+}
+
++ (void)updateCOPPAStatus:(BOOL)optIn {
+  [VunglePrivacySettings setCOPPAStatus:optIn];
```
joshh-devrel, line 29: What if the publisher hasn't specified COPPA value?
https://developers.google.com/admob/ios/targeting#child-directed_setting

Should we just not set the COPPA value in the Vungle SDK?

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
```diff
@@ -50,14 +48,21 @@ - (nonnull instancetype)initWithGADMAdNetworkConnector:(nonnull id<GADMAdNetwork
...
+
 - (void)getBannerWithSize:(GADAdSize)adSize {
   id<GADMAdNetworkConnector> strongConnector = _connector;
   id<GADMAdNetworkAdapter> strongAdapter = _adapter;
   if (!strongConnector || !strongAdapter) {
     return;
   }
 
-  _bannerSize = GADMAdapterVungleAdSizeForAdSize(adSize);
+  _bannerSize = [self filterValidAdSizes:adSize];
```
joshh-devrel, line 65: Why do we have the below for bidding but not waterfall? Can we just include `vungleAdSizeForAdSize` in utils and use it for both bidding and waterfall?
`_bannerSize = [self vungleAdSizeForAdSize:adConfiguration.adSize];`

Also the method name with `filter` sounds odd to me.

adapters/Vungle/Public/Headers/VungleAdNetworkExtras.h
stanley-vungle: Yes. Those properties are no longer supported anymore without any substitutions.

adapters/Vungle/Public/Headers/VungleRouterConsent.h
stanley-vungle: It can be anytime. But for it to take effect, it needs to happen before the ad request is made.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
stanley-vungle: This callback will be called when the banner ad is presented. We do not have an equivalent callback for when the modal is presented after the banner ad is clicked.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
stanley-vungle: Done.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleInterstitial.m
stanley-vungle: No. It should be very close. The DidPresent callback is called when the view is successfully attached, and the impression callback is called when our ad plays successfully and sends a signal to the Vungle SDK to trigger.

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
stanley-vungle: Done.

adapters/Vungle/VungleAdapter/VungleRouterConsent.m
stanley-vungle: I will remove this wrapper around the Vungle COPPA. There was a previous comment somewhere that the publisher should use the Google SDK API to set the COPPA flag and the adapter would attempt to pull the childDirectedTreatment flag from the Google SDK on every load request.
https://github.com/Vungle/googleads-mobile-ios-mediation/blob/7.0.x_pre_release/adapters/Vungle/VungleAdapter/GADMediationAdapterVungle.m#L111

But I just moved the method away from the public header at the time.

adapters/Vungle/VungleAdapter/GADMAdapterVungleBanner.m
stanley-vungle: Ok. Done.

adapters/Vungle/VungleAdapter/Bidding/GADMediationVungleBanner.m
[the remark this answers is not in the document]
stanley-vungle: Moved this method to utils to follow review from joshh.

joshh-devrel:
@stanley-vungle Can you rebase this pull request as we just finished the rebrand to Liftoff Monetize?

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/VungleRouterConsent.m
```diff
@@ -13,23 +13,20 @@
...
+}
+
++ (void)updateCCPAStatus:(BOOL)optIn {
+  [VunglePrivacySettings setCCPAStatus:optIn];
 }
 
-+ (VungleConsentStatus)getConsentStatus {
-  return currentConsentStatus;
++ (void)updateIDFVStatus:(BOOL)optIn {
+  [VunglePrivacySettings setPublishIdfv:optIn];
```
joshuajplagonera, line 29: Is there a reason this class is still here? I think this class can be removed.
It looks like the publishers can directly call these `VunglePrivacySettings` methods statically and we can update the docs to clarify they need to be set before ad request time.

stanley-vungle: @joshuajplagonera Another reason why it was wanted was to avoid less direct interaction with our SDK so if the API changes, the migration for the publisher would be easier.

I got the approval to remove this if this is a blocker for the certification.

joshuajplagonera: Thanks for the explanation. This can be kept. :)

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,92 @@ - (void)requestBannerAd {
...
-  return [_adConfiguration bidResponse];
-}
-
-- (GADAdSize)bannerAdSize {
-  return _bannerSize;
+  _bannerAd = [[VungleBanner alloc]
+      initWithPlacementId:self.desiredPlacement
+                     size:GADMAdapterVungleConvertGADAdSizeToBannerSize(_bannerSize)];
+  _bannerAd.delegate = self;
+  VungleAdsExtras *extras = [[VungleAdsExtras alloc] init];
```
joshuajplagonera, line 120: We're having build issues where `VungleAdsExtras` cannot be found:

```
Use of undeclared identifier 'VungleAdsExtras'
```

stanley-vungle: This was support for the watermark. I thought that it was going to be released soon on our end so I included this back in. Let me roll this commit back.

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleBanner.m
```diff
@@ -101,154 +98,89 @@ - (void)requestBannerAd {
...
 }
 
-- (void)rewardUser {
-  // Do nothing.
-}
+#pragma mark - GADMAdapterVungleDelegate
 
-- (void)didShowAd {
-  // Do nothing.
+- (void)initialized:(BOOL)isSuccess error:(nullable NSError *)error {
```
joshh-devrel, line 178: Can you remind me again when would/should this `initialized` callback be called? Do we really need this callback?

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMAdapterVungleUtils.m
```diff
@@ -137,13 +72,24 @@ GADAdSize GADMAdapterVungleAdSizeForAdSize(GADAdSize adSize) {
...
 }
 
+BannerSize GADMAdapterVungleConvertGADAdSizeToBannerSize(GADAdSize adSize) {
+  if (GADAdSizeEqualToSize(adSize, GADAdSizeMediumRectangle)) {
+    return BannerSizeMrec;
+  }
+  if (adSize.size.height == GADAdSizeLeaderboard.size.height) {
+    return BannerSizeLeaderboard;
+  }
+  // Height is 50.
```
joshh-devrel, line 86: Can you clarify why is that the case?

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMediationAdapterVungle.h
```diff
@@ -18,16 +18,16 @@
 typedef NS_ENUM(NSInteger, GADMAdapterVungleErrorCode) {
   /// Missing server parameters.
   GADMAdapterVungleErrorInvalidServerParameters = 101,
-  /// An ad is already loaded for this network configuration. Vungle SDK cannot load
+  /// An ad is already loaded for this network configuration. Liftoff Monetize SDK cannot load
```
joshh-devrel, line 21: It's still "Vungle SDK" in [Liftoff's documentation](https://support.vungle.com/hc/en-us/articles/360002925791), but I will defer to Liftoff team on whether it should be called Vungle or Liftoff Monetize SDK here.

joshh-devrel: The same applies to the other places.

adapters/LiftoffMonetize/Public/Headers/VungleAdNetworkExtras.h
```diff
@@ -24,25 +24,26 @@
...
+@property(nonatomic, copy) NSArray<NSString *> *_Nullable allPlacements DEPRECATED_ATTRIBUTE;
 
-@property(nonatomic, copy, readonly) NSString *_Nonnull UUID;
+@property(nonatomic, copy) NSNumber *_Nullable orientations DEPRECATED_ATTRIBUTE;
 
-@property (nonatomic, readonly, assign) BOOL muteIsSet;
+@property(nonatomic, copy, readonly) NSString *_Nonnull UUID DEPRECATED_ATTRIBUTE;
 
-@property (nonatomic, assign) NSInteger nativeAdOptionPosition;
+@property (nonatomic, readonly, assign) BOOL muteIsSet DEPRECATED_ATTRIBUTE;
```
joshh-devrel, line 47: Seems like these attributes aren't used anywhere? Can we remove them?

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMediationAdapterVungle.m
```diff
@@ -42,11 +42,13 @@ @implementation GADMediationAdapterVungle {
 
 + (void)setUpWithConfiguration:(nonnull GADMediationServerConfiguration *)configuration
              completionHandler:(nonnull GADMediationAdapterSetUpCompletionBlock)completionHandler {
+  // Gather all app ids supplied by the server configuration
   NSMutableSet *applicationIDs = [[NSMutableSet alloc] init];
-
   for (GADMediationCredentials *cred in configuration.credentials) {
     NSString *appID = cred.settings[GADMAdapterVungleApplicationID];
-    GADMAdapterVungleMutableSetAddObject(applicationIDs, appID);
+    if (appID) {
```
joshh-devrel, line 49: Just want to confirm if we need to do `if (appID.length)`.

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMAdapterVungleBanner.m
```diff
@@ -71,139 +76,93 @@ - (void)getBannerWithSize:(GADAdSize)adSize {
...
-  }
+#pragma mark - VungleBannerDelegate
 
-  if (self.isAdLoaded) {
-    // Already invoked an ad load callback.
-    return;
-  }
-  self.isAdLoaded = YES;
-  UIView *bannerView = [[UIView alloc]
+- (void)bannerAdDidLoad:(VungleBanner *)banner {
```
joshh-devrel, line 110: Should indicate nullability for the input parameter.

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleBanner.m
joshh-devrel: @stanley-vungle Just to confirm - so watermark for Vungle bidding ads isn't supported yet, correct? Would you know when will it be supported?

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/VungleRouterConsent.m
stanley-vungle: Ok. I'll remove the class.

The consent values should be set prior to the requesting ads of the Vungle SDK (through the GMA SDK). Currently our SDK does not utilize any identifiers for the initialization but it will be passed when the ad request is made.

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleBanner.m
stanley-vungle: This `initialized` callback is called if the Vungle SDK wasn't initialized yet. It is triggered as a form of lazy initialization. If the publisher's app initializes the Vungle SDK before the load request, it will never be hit. If the Vungle SDK hasn't been initialized, the load request will first attempt initialization and on the callback, it will trigger this.

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMAdapterVungleUtils.m
stanley-vungle: It is the only supported size from Vungle. Our SDK will try to fit the banner in the view, but the true height size of the assets is 50 px

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMediationAdapterVungle.h
stanley-vungle: @joshh-devrel In the [re-branding commit](https://github.com/googleads/googleads-mobile-ios-mediation/commit/1bee38f642f6b0cbbf9bbdd0ba66899be3d60a51), almost all comments were changed from `Vungle` to `Liftoff Monetize`. I thought that these cases were missed so I changed it. I'll remove the "SDK" part of the comment.

adapters/LiftoffMonetize/Public/Headers/VungleAdNetworkExtras.h
stanley-vungle: Ok :+1:

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMediationAdapterVungle.m
stanley-vungle: Our SDK has a check internally to see if the string is empty and send an error, but I think your appID.length would handle more integration mishaps. I'll change it.

PR #396, opened 2023-02-06
Update Unity iOS adapter to use new COPPA method

joshh-devrel:
Update Unity iOS adapter to use new COPPA method


joshh-devrel:
Closed this PR, use https://github.com/googleads/googleads-mobile-ios-mediation/pull/400 instead.

PR #411, opened 2023-04-17
Inmobi Adapter init call on main thread

kshtgarg21:


adapters/InMobi/InMobiAdapter/GADMAdapterInMobiInitializer.m
```diff
@@ -61,28 +61,43 @@ - (void)initializeWithAccountID:(nonnull NSString *)accountID
...
-        for (GADMAdapterInMobiInitCompletionHandler completionHandler in strongSelf
-                 ->_completionHandlers) {
-          completionHandler(error);
-        }
-      }];
+    GADMAdapterInMobiInitializer *__weak weakSelf = self;
+
+    if ([NSThread isMainThread]) {
+        [self initializeInMobiSDK:accountID completionHandler:completionHandler];
+    } else {
```
joshuajplagonera, line 68: I think this check is not necessary. 
Calling `dispatch_async(dispatch_get_main_queue() ...` should already add this to the queue on the main thread, even if its already in the main thread.

kshtgarg21: without this check init call would become async if already on main thread. 
We want ii to remain sync if already on main thread

PR #415, opened 2023-05-08
Added support for the arm64 simulator architecture in LiftoffMonetize…

stsandro:
… Adapter

This should make it possible to test on the simulator on Apple Silicon Macs.


stsandro:
With 5ce1f9981def54af7df470371f209a1e72ae1fde merged in this is no longer necessary.

Thanks 🙏

PR #427, opened 2023-08-08
Mandate information icon image to serve native ad. If it fails to load, then this is considered as native ad load failure now.

hikaruwatanabe-eng:
Mandate information icon image to serve native ad. If it fails to load, then this is considered as native ad load failure now.


hidetoshi-muta (approved):
Thank you for quick PR.
LGTM.

PR #442, opened 2023-11-27
[Liftoff Monetize] 7.2.0.0 - Update error objects to use the Liftoff errors where applicable

stanley-vungle:
The Liftoff error codes has been added to public documentation.

Please update Google's public documentation for Liftoff SDK to refer to this page for the error codes:
https://support.vungle.com/hc/en-us/articles/15661036969627

Thanks.

### DIFF
--- a/adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleInterstitial.m
+++ b/adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleInterstitial.m
@@ -50,8 +50,7 @@
   if (self) {
     _adConfiguration = adConfiguration;
     self.desiredPlacement =
-        [GADMAdapterVungleUtils findPlacement:adConfiguration.credentials.settings
-                                networkExtras:adConfiguration.extras];
+        [GADMAdapterVungleUtils findPlacement:adConfiguration.credentials.settings];
 
     __block atomic_flag adLoadHandlerCalled = ATOMIC_FLAG_INIT;
     __block GADMediationInterstitialLoadCompletionHandler origAdLoadHandler =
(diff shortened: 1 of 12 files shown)


adapters/LiftoffMonetize/LiftoffMonetizeAdapter/GADMediationAdapterVungle.h
```diff
@@ -18,17 +18,8 @@
 typedef NS_ENUM(NSInteger, GADMAdapterVungleErrorCode) {
```
stanley-vungle, line 18: The other error codes are handled within the SDK. The two remaining ones issues when retrieving valid information or object from the adapter from the Google SDK

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleInterstitial.m
```diff
@@ -75,12 +74,6 @@ @implementation GADMediationVungleInterstitial {
 }
 
 - (void)requestInterstitialAd {
-  if (!self.desiredPlacement.length) {
```
hikaruwatanabe-eng, line 78: Is there any reason we want to remove it? Looks like it still can be  @""

stanley-vungle: It's ok if it is an empty string. The SDK has been updated to be able to handle these cases and will properly return a failed to load callback when the load is requested.

hikaruwatanabe-eng: Ah okay! Thank you for the explanation.

adapters/LiftoffMonetize/LiftoffMonetizeAdapter/Bidding/GADMediationVungleRewardedAd.m
```diff
@@ -92,6 +85,13 @@ - (void)loadRewardedAd {
   VungleAdsExtras *extras = [[VungleAdsExtras alloc] init];
   [extras setWithWatermark:[_adConfiguration.watermark base64EncodedStringWithOptions:0]];
   [_rewardedAd setWithExtras:extras];
+    if ([_adConfiguration extras] &&
+        [[_adConfiguration extras] isKindOfClass:[VungleAdNetworkExtras class]]) {
+      VungleAdNetworkExtras *extras = (VungleAdNetworkExtras *)[_adConfiguration extras];
```
hikaruwatanabe-eng: I think this should be named networkExtras or should be refactored into a separate function/method like `findPlacement:`. One can get confused with extras declared at line 92/85.

stanley-vungle: Ok. Updated

PR #459, opened 2024-06-11
Added support for bidding native ads.

hikaruwatanabe-eng:
Added support for bidding native ads.

### DIFF
--- a/adapters/Line/LineAdapter/GADMediationAdapterLineNativeAdLoader.m
+++ b/adapters/Line/LineAdapter/GADMediationAdapterLineNativeAdLoader.m
@@ -69,15 +69,10 @@ static NSUInteger GADMediationAdapterLineImageAssetLoadingTimeoutInSeconds = 10;
     return;
   }
 
-  NSString *slotID = GADMediationAdapterLineSlotID(_adConfiguration.credentials, &error);
-  if (error) {
-    [self callCompletionHandlerIfNeededWithAd:nil error:error];
-    return;
-  }
-
   _shouldLoadAdImages = YES;
   BOOL shouldEnableSound = !GADMobileAds.sharedInstance.applicationMuted;
   NSUInteger numberOfImageAdLoaderOptions = 0;
+  NSUInteger numberOfVideoAdLoaderOptions = 0;
   for (GADAdLoaderOptions *loaderOptions in _adConfiguration.options) {
     if ([loaderOptions isKindOfClass:[GADNativeAdImageAdLoaderOptions class]]) {
       GADNativeAdImageAdLoaderOptions *imageOptions =
@@ -89,6 +84,7 @@ static NSUInteger GADMediationAdapterLineImageAssetLoadingTimeoutInSeconds = 10;
     if ([loaderOptions isKindOfClass:[GADVideoOptions class]]) {
       GADVideoOptions *videoOptions = (GADVideoOptions *)loaderOptions;
       shouldEnableSound = !videoOptions.startMuted;
+      numberOfVideoAdLoaderOptions += 1;
     }
   }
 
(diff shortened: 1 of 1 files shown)


phenan:
[question]

Is it okay for `GADMediationAdapterLineNativeAdLoader` API to ignore `extras.nativeAdVideoWidth` if it is specified?
If you need to consider `extras.nativeAdVideoWidth`, I think you need to resize the view obtained with `getAdMainView` appropriately.

hikaruwatanabe-eng:
> [question]
> 
> Is it okay for `GADMediationAdapterLineNativeAdLoader` API to ignore `extras.nativeAdVideoWidth` if it is specified? If you need to consider `extras.nativeAdVideoWidth`, I think you need to resize the view obtained with `getAdMainView` appropriately.
> 
> `GADMediationAdapterLineNativeAdLoader` API は `extras.nativeAdVideoWidth` が指定されていた場合にそれを無視しても良いのでしょうか？ もし `extras.nativeAdVideoWidth` を考慮する必要がある場合は、`getAdMainView` で取得したビューを適切にリサイズする必要があると思います

That was my question too. Thank you the clarification! I will make an update.

hikaruwatanabe-eng:
Updated the code. The extras.nativeAdViewoWidth was added before because FiveAd needed the parameter in the older API.

```
 [[FADNative alloc] initWithSlotId:slotID videoViewWidth:extras.nativeAdVideoWidth];
```

But now the ad loader doesn't take it, I just resized the native ad's main view as suggested. If you think it needs to be done other way, please let me know.

adapters/Line/LineAdapter/GADMediationAdapterLineNativeAdLoader.m
```diff
@@ -103,16 +99,87 @@ - (void)loadAd {
...
+                 [strongSelf callCompletionHandlerIfNeededWithAd:nil error:error];
+                 return;
+               }
+
+               GADMediationAdapterLineExtras *extras = strongSelf->_adConfiguration.extras;
+               if (!extras) {
+                 UIView *nativeAdMainView = [nativeAd getAdMainView];
+                 CGRect newFrame = nativeAdMainView.frame;
+                 newFrame.size.width = extras.nativeAdVideoWidth;
+                 nativeAdMainView.frame = newFrame;
```
phenan, line 173: I think you should also set height to keep the aspect ratio as follows:

```objc
CGRect currentSize = nativeAdMainView.frame.size;
CGRect newFrame = nativeAdMainView.frame;
newFrame.size = CGSizeMake(extras.nativeAdVideoWidth, extras.nativeAdVideoWidth * currentSize.height / currentSize.width);
```

phenan (commented):
The new API does not have width parameter as you know, we don't know this design choice is good or bad.

From Google's perspective, do you think it's better to take the width explicitly?

If it is better to take width, it is possible to release a version of the SDK with improved API within a few days.

---

新しい AdLoader API は width を取らない形式になっていますが、このデザインが良いかどうかはまだ悩んでいる段階です。

Google からみて、width は引数に取るべきだと思うでしょうか？

もしそうであれば、数日以内に width を引数に取るバージョンの SDK をリリースする用意があります。

adapters/Line/LineAdapter/GADMediationAdapterLineNativeAdLoader.m
```diff
@@ -103,16 +99,87 @@ - (void)loadAd {
...
+                 GADMediationAdapterLineLog(
+                     @"FiveAd SDK failed to load a bidding native ad. The FiveAd error code: %ld.",
+                     adLoadError.code);
+                 error = GADMediationAdapterLineErrorWithFiveAdErrorCode(adLoadError.code);
+                 [strongSelf callCompletionHandlerIfNeededWithAd:nil error:error];
+                 return;
+               }
+
+               GADMediationAdapterLineExtras *extras = strongSelf->_adConfiguration.extras;
+               if (!extras) {
```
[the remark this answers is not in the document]
hikaruwatanabe-eng: oops sorry will update this.

adapters/Line/LineAdapter/GADMediationAdapterLineNativeAdLoader.m
hikaruwatanabe-eng: okay, done!

hikaruwatanabe-eng:
> The new API does not have width parameter as you know, we don't know this design choice is good or bad.
> 
> From Google's perspective, do you think it's better to take the width explicitly?
> 
> If it is better to take width parameter, it is possible to release a version of the SDK with improved API within a few days.
> 
> 新しい AdLoader API は width を取らない形式になっていますが、このデザインが良いかどうかはまだ悩んでいる段階です。
> 
> Google からみて、width は引数に取るべきだと思うでしょうか？
> 
> もしそうであれば、数日以内に width を引数に取るバージョンの SDK をリリースする用意があります。

Either works, but I think it would be better for LINE SDK to accept the width parameter so that in the future, if this API is used by anyone other than the adapter, LINE would have control over how to set the width. This way, the ad sizing logic would most likely be consistent across different API consumers.

phenan (approved):

phenan:
> Either works, but I think it would be better for LINE SDK to accept the width parameter so that in the future, if this API is used by anyone other than the adapter, LINE would have control over how to set the width.

Thank you for your kindly response.
I will reconsider the API within my team.

We will contact you again if we change the API.
(We do not intend to make any breaking changes to the current API)

PR #467, opened 2024-09-03
7.4.1 Vungle SDK release

ManojBudumuru-vungle:
Optimization in SDK initialization
Update OMSDK to version 1.5.0
Bug fixes and stability improvements

### DIFF
--- a/adapters/LiftoffMonetize/Public/Headers/VungleAdNetworkExtras.h
+++ b/adapters/LiftoffMonetize/Public/Headers/VungleAdNetworkExtras.h
@@ -26,9 +26,10 @@
 @property(nonatomic, copy) NSString *_Nullable playingPlacement;
 
 /*!
- * @brief NSInteger that will be passed to alter the privacy icon position for native ads.
- * @discussion Optional. topLeft = 1, topRight = 2, bottomLeft = 3, bottomRight = 4
+ * @brief GAAdChoicesPosition enum that will be passed to alter the privacy icon position for
+ * native ads.
+ * @discussion Optional. topRight = 0, topLeft = 1, bottomRight = 2, bottomLeft = 3
  */
-@property(nonatomic, assign) NSInteger nativeAdOptionPosition;
+@property(nonatomic, assign) GADAdChoicesPosition nativeAdOptionPosition;
 
 @end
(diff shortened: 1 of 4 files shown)


adapters/LiftoffMonetize/Public/Headers/VungleAdNetworkExtras.h
```diff
@@ -26,9 +26,10 @@
 @property(nonatomic, copy) NSString *_Nullable playingPlacement;
 
 /*!
- * @brief NSInteger that will be passed to alter the privacy icon position for native ads.
- * @discussion Optional. topLeft = 1, topRight = 2, bottomLeft = 3, bottomRight = 4
+ * @brief GADAdChoicesPosition enum that will be passed to alter the privacy icon position for
+ * native ads.
+ * @discussion Optional. topRight = 0, topLeft = 1, bottomRight = 2, bottomLeft = 3
```
thanvirm, line 31: Wouldn't changing this break existing publishers who use this extra?

Since GADAdChoicesPosition is really an NSInteger underneath, the publishers wouldn't face any compilation error. But, I think there will be unexpected behavior change in the app because of this change.

ManojBudumuru-vungle: @thanvirm We had to make sure our positioning is consistent with Googles positioning. liftoff enums are different. 
We had a publisher complained about this. So we did this change to match google. This will make sure that the pubs will have the proper position.

ManojBudumuru-vungle: @thanvirm here is the change log for this change.
NativeAdOptionPosition has been updated to align with GADAdChoicesPosition. Use GADAdChoicesPosition enum values when specifying the privacy icon position (Google documentation reference [here](https://developers.google.com/admob/ios/api/reference/Enums/GADAdChoicesPosition))

dose this looks good.? should i add this once this adapter version is released or can u add this point? Thanks

thanvirm: Thank you Manoj. I will add this to the change log.

PR #475, opened 2024-12-10
#Moloco #Adapter Set isAgeRestrictedUser bit on Moloco.

thanvirm:
#Moloco #Adapter Set isAgeRestrictedUser bit on Moloco.


ajithvl (approved):
Looks good to me

PR #482, opened 2025-03-13
Added support for COPPA flag.

hikaruwatanabe-eng:
Added support for COPPA flag.


Mintegral-Lee (approved):
I have reviewed all commits and found no issues.

PR #492, opened 2025-04-09
Inmmobi int rewarded- Added supoort for GADMAdapterInMobiInterstitialRewardedAd

kshtgarg21:


thanvirm:
I believe this is obsolete since https://github.com/googleads/googleads-mobile-ios-mediation/pull/493 got merged. Thanks!

PR #1035, opened 2026-05-07
Fix RTB banner size propagation and native body in BidMachine adapter

AssasinSC:
## Summary                                                                                                                  
                                                                                                                              
  This PR fixes two bugs in the BidMachine adapter:                                                                           
                                                                                                                              
  **1. RTB banner size not propagated**                                                                                       
                                                                                                                              
  `collectSignals` and `loadRTBBannerAd` were using a hardcoded `.banner` (320×50) placement format regardless of the         
  requested ad size. As a result, MREC (300×250) and leaderboard (728×90) RTB requests were always sent with the wrong
  placement format to BidMachine.                                                                                             
                                                     
  Fix: extract `AdSize` from `RTBRequestParameters`/`MediationBannerAdConfiguration` and map it to the correct                
  `PlacementFormat` via `AdSize.toPlacementFormat()`.
                                                                                                                              
  **2. Native ad body returning NSObject description**
                                         
  `NaitveAdProxy.body` was returning `nativeAd.description` (which resolves to `NSObject.description` — a memory address      
  string) instead of `nativeAd.body`.
                                                                                                                              
  Fix: changed to `nativeAd.body`.                                                                                            
                                         
  ## Testing                                                                                                                  
                                                     
  - Added unit tests for MREC and leaderboard size handling in `collectSignals` and `loadRTBBannerAd`                         
  - Added unit tests for unsupported banner size rejection
  - All 78 existing tests pass


adapters/BidMachine/BidMachineAdapter/BidMachineClient.swift
```diff
@@ -133,37 +133,22 @@ final class BidMachineClientImpl: NSObject, BidMachineClient {
...
   }
 
   func loadRTBBannerAd(
     with bidResponse: String,
+    size: AdSize,
     delegate: BidMachineAdDelegate,
     watermark: String,
     completionHandler: @escaping (NSError?) -> Void
   ) throws {
+    let bannerFormat = try size.toPlacementFormat()
```
thanvirm: With this change, RTB banner ad load will start failing for non-standard (i.e. adaptive) banner sizes. Note: The size that's passed to loadRTBBannerAd is the size that the publisher requested and not the size BidMachine may have responded with in the bid response.

I would recommend not having a size check on the adapter when loading banner ads for RTB. The publisher-requested size is passed as part of the bid request during auction and  the BidMachine bidder can decide at that stage whether to bid or not. Note: Adapter's loadRTBBannerAd occurs only after auction has completed and BidMachine has won the auction. (Reference: https://developers.google.com/admob/ios/open-bidding-adapter#workflow_of_a_bidding_adapter).

Also, here is how adaptive banner ad request will be represented in the bid request: https://developers.google.com/authorized-buyers/rtb/adaptive-banner

We can discuss this further during our team sync if needed.

thanvirm (changes requested):

adapters/BidMachine/BidMachineAdapter/BidMachineClient.swift
staskochkin: Thanks for the careful review! Let me explain why we need the size propagated to `loadRTBBannerAd` and why adaptive banners are not impacted.

**1. BidMachine's RTB flow is size-aware end-to-end, not just at auction time**

BidMachine's SDK keys its on-device bid-response cache and its load-side `BannerRequest` builder by `PlacementFormat`. The placement format must be consistent across the full lifecycle:

- `collectSignals` -> builds the OpenRTB `imp.banner` (w/h + `format[]`) and stores any cached state under that format.
- `loadRTBBannerAd` -> resolves the cached payload and constructs the renderer for the same format.

With the previous hardcoded `.banner` (320x50):
- A publisher MREC (300x250) request would go out as 320x50 in the bid request, so the BidMachine bidder either wouldn't bid or would respond with a 320x50 creative.
- Even when a response came back, the load-side `BannerRequest` was built for 320x50, so the creative was handed to Google's `GADBannerView` sized for MREC/leaderboard. The dimension mismatch is what caused the "failed to present" behavior on MREC and 728x90 that we were seeing in production.

Passing the publisher-requested size to both call sites is what realigns the bid request, the cached response, and the renderer.

**2. Adaptive banners are not rejected - they are bucketed**

`AdSize.toPlacementFormat()` is an approximation, not a strict equality check. Any incoming `AdSize` (including adaptive sizes where Google passes a width with an adaptive height) is mapped to the closest standard `PlacementFormat` BidMachine supports: `.banner`, `.mrec`, or `.leaderboard`. This matches how BidMachine handles adaptive in its direct (non-mediated) SDK integrations and is consistent with the adaptive-banner bid-request spec you linked: the bid request carries the requested w/h plus `format[]` alternatives, and the bidder returns a creative sized to one of the standard buckets.

The one path that does throw is genuinely-unsupported input (e.g. zero/negative dimensions). The unit test in the PR that asserts rejection is for that sanity-guard case, not for adaptive sizes.

**3. Re: "the bidder can decide at auction"**

Agreed for the auction decision — but the load path on the device still needs the size to:
- Look up the correct cached bid response,
- Build the size-aware `BannerRequest` for the load,
- Provide correct dimensions to BidMachine's renderer for OMSDK viewability/impression reporting (which keys off banner size).

If `loadRTBBannerAd` always passes `.banner`, the bid can be won but the on-device load silently mismatches and ad presentation fails — which is the bug this PR is fixing.

Happy to walk through this at the team sync. We've verified end-to-end on MREC, 320x50, and 728x90 with the proposed change and all three render correctly; on the previous code path MREC and 728x90 were failing to present.

staskochkin: Hello @thanvirm, 
We have reworked the logic for the size mapping. For waterfall integration, we leave `closestValidSizeForAdSizes` from Google's utilities. For bidding, we use a new custom method to map size. Could you please take a look? 
Thanks in advance

adapters/BidMachine/BidMachineAdapter/BidMachineClient.swift
```diff
@@ -384,3 +375,36 @@ extension GoogleMobileAds.AdFormat {
...
+    } else {
+      throw BidMachineAdapterError(
+        errorCode: .unsupportedBannerSize, description: "Unsupported banner size.")
+    }
+  }
+
+  /// Maps an ad size to a BidMachine placement format for bidding requests.
+  /// Always returns one of the three supported placements based on dimensions.
+  fileprivate func toBiddingPlacementFormat() -> PlacementFormat {
+    if self.size.height >= 200 { return .banner300x250 }
```
thanvirm: if the requested height here is say 225, then I think BidMachine would display a 300x250 ad. I think that's not good since the displayed ad may cover the app's content.

I would suggest rewriting this function as:

```
If width >= 728 & height >= 90:
	return 728x90
If width >= 300 & height >= 250:
	return 300x250
If with >= 320 & height >= 50:
	return 320x50

// if the requested size is inline adaptive with no height restriction, the height will be specified as 0.
If height == 0:
	If width >= 728:
		return 728x90
	else if width >= 320:
		return 320x50
	else if width >= 300:
		return 300x250

return error
```

What do you think?

thanvirm (changes requested):

adapters/BidMachine/BidMachineAdapter/BidMachineClient.swift
staskochkin: BidMachine SDK clips a creative's content to Banner View bounds, so the banner won't overlap an app's content. However, the requested changes are still valid, and I'll implement them. Thank you!

staskochkin: @thanvirm I reworked this utility method. In your proposed code, there was a gap. For inline adaptive banners with a width of 300..<320, the MREC ad size is always requested.  This can cause a request MREC for actual sizes like 300x50, etc. 
In the current solution, we won't use MREC for inline adaptive banners at all, and will request 320x50 instead. Method might returns error only in cases when requested container is less than we can actually fill (I checked size in the reporting tool)

PR #1116, opened 2026-05-29
[Liftoff Monetize] 7.7.5.1

ashinagawa:
- Removed the `canPlayAd` check for AppOpen ads to let them play immediately. This allows the Liftoff SDK to handle the case when an ad is not ready, trigger the `failedToPlay` callback to adapter, and log the event properly (for Liftoff side).


komal188d:
Hello @ashinagawa ,

Seems like the PR is incomplete. Can you please make changes to the podspec and test cases(wherever required), so that it doesn't fail.

Thank you!!
Komal

ashinagawa:
@komal188d 
Thanks for quick review, and reply.
I fixed the unit tests and updated the version properly.

Thanks again,
-aki

PR #1183, opened 2026-06-25
[Moloco] Add adaptive banner support and update to Moloco SDK 4.8.0

seoyomi:
## Summary
Adds **adaptive banner** support to the Moloco iOS mediation adapter and bumps the `MolocoSDKiOS` dependency to **4.8.0** (the release that introduced the adaptive banner API). Mirrors the Android adapter change in googleads/googleads-mobile-android-mediation#1344.

Previously `BannerAdLoader` selected a Moloco banner by branching on a couple of fixed `AdSize` values (MREC vs standard). This PR replaces that with the unified `Moloco.shared.createMolocoBanner(params:size:viewController:)` API plus a size-mapping function, so the adapter can serve inline- and anchored-adaptive banners in addition to the fixed sizes.

## Changes
- **`BannerAdLoader.swift`** — add `molocoBannerAdSize(from:)`; `loadAd()` resolves `adConfiguration.adSize` and calls the unified factory.
- **`MolocoBannerFactory.swift`** — replace `createBanner` / `createMREC` with a single size-aware `createBanner(for:size:delegate:watermarkData:)`.
- **`MolocoSdkImpl.swift`** — implement it via `Moloco.shared.createMolocoBanner(...)`.
- **podspec** — `MolocoSDKiOS` 4.7.0 → 4.8.0.

## Size mapping (mirrors #1344, adapted for iOS)
1. exact `AdSizeBanner` → `.standard`, `AdSizeMediumRectangle` → `.mrec`
2. height matches `currentOrientationAnchoredAdaptiveBanner(width:)` → `.anchoredAdaptive(width:)`
3. fallback (inline adaptive, leaderboard, custom width) → `.inlineAdaptive(width:)`

iOS differences from Android: Google's inline `height == 0` signal doesn't apply on iOS (inline is the fallback), Moloco iOS has no tablet/leaderboard fixed size, and no `Context` threading is needed (the iOS anchored API takes only `width`).

## Note for reviewers
`currentOrientationAnchoredAdaptiveBanner(width:)` is deprecated in GMA 13.x (in favor of the `large…` variants). Moloco's SDK doesn't expose an anchored adaptive size for comparison, so the GMA API is the only way to distinguish anchored from inline here. Happy to switch to the `large…` variant, or to drop the anchored/inline split (Moloco renders them identically), per your preference.

### DIFF
--- a/AdapterUnitTests/AdapterUnitTests.xcodeproj/project.pbxproj
+++ b/AdapterUnitTests/AdapterUnitTests.xcodeproj/project.pbxproj
@@ -48,6 +48,7 @@
 		856602ED2C642DE70068D786 /* FakeMolocoRewarded.swift in Sources */ = {isa = PBXBuildFile; fileRef = 856602EC2C642DE70068D786 /* FakeMolocoRewarded.swift */; };
 		8595C0C42C59A82B00B46FD2 /* MolocoRewardedAdTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8595C0C32C59A82B00B46FD2 /* MolocoRewardedAdTest.swift */; };
 		85A57D6F2CA4B640009FC260 /* MolocoBannerAdTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 85A57D6E2CA4B640009FC260 /* MolocoBannerAdTest.swift */; };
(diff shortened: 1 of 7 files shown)


LTPhantom:
Hi and thanks for the contribution! Could you help us by first rebasing the PR since there seems to be some branch conflicts and then by adding some unit tests for the new behavior if needed?

komal188d:
Hello @seoyomi ,

I see you have updated the code, but the test cases aren't updated. Can you please update and I can review them as well.
Thank you!!

seoyomi:
Hi @LTPhantom @komal188d, I've added unit tests. Could you guys review it and let me know if it's sufficient? Thank you!

PR #1216, opened 2026-07-16
Fix UnityAds adapter to fail signal collection on empty bidding token

vishaldhiman-unity:
When Unity Ads returns a null or empty bidding token, the adapter was passing it through as a successful signal (an empty string). This effectively looked like a no fill downstream instead of a real failure.

This change routes null or empty tokens to the completion handler as a failure, with a new error code, GADMAdapterUnityErrorEmptyBiddingToken. Updated the existing nil token test and added a test for the empty token case.


vishaldhiman-unity:
Closing in favor of #1220, which recreates this change as a single commit with correct authorship.

PR #1220, opened 2026-07-17
Fix UnityAds adapter to fail signal collection on empty bidding token

vishaldhiman-unity:
When Unity Ads returns a null or empty bidding token, the adapter was passing it through as a successful signal (an empty string). This effectively looked like a no fill downstream instead of a real failure.

This change routes null or empty tokens to the completion handler as a failure, with a new error code, GADMAdapterUnityErrorEmptyBiddingToken. Updated the existing nil token test and added a test for the empty token case.

Supersedes #1216 (recreated as a single commit to fix commit attribution).

### DIFF
--- a/AdapterUnitTests/UnityAdapterTests/AUTUnityAdapterTests.m
+++ b/AdapterUnitTests/UnityAdapterTests/AUTUnityAdapterTests.m
@@ -236,8 +236,41 @@
   [adapter
       collectSignalsForRequestParameters:params
                        completionHandler:^(NSString *_Nullable signals, NSError *_Nullable error) {
-                         XCTAssertNil(error);
-                         XCTAssertEqualObjects(signals, @"");
+                         XCTAssertNil(signals);
+                         XCTAssertEqual(error.code, GADMAdapterUnityErrorEmptyBiddingToken);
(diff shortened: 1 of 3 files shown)


adapters/Unity/UnityAdapter.xcodeproj/project.pbxproj
```diff
@@ -527,6 +553,28 @@
...
 		};
+		68E97D7604FE6E5FD9DB9F4C /* [CP] Check Pods Manifest.lock */ = {
+			isa = PBXShellScriptBuildPhase;
+			buildActionMask = 2147483647;
+			files = (
+			);
+			inputFileListPaths = (
+			);
+			inputPaths = (
+				"${PODS_PODFILE_DIR_PATH}/Podfile.lock",
```
ericleich: Can you run "pod deintegrate" and reupload so this diff does not show up in this file?

vishaldhiman-unity:
Ran "pod deintegrate" and force-pushed a cleaned-up commit.

adapters/Unity/UnityAdapter.xcodeproj/project.pbxproj
vishaldhiman-unity: Addressed. Thank you!

PR #1249, opened 2026-08-05
 Add placement ID support to BidMachine iOS mediation adapter.

AssasinSC:


adapters/BidMachine/BidMachineAdapter/Util.swift
```diff
@@ -65,6 +66,29 @@ final class Util {
...
+  /// - Returns: A placement ID from the configuration, or nil if the configuration contains none.
+  static func placementId(from config: MediationAdConfiguration) -> String? {
+    return config.credentials.settings[MediationConfigurationSettingKey.placementId.rawValue]
+      as? String
+  }
+
+  /// Retrieves a placement ID from the provided RTB parameters.
+  ///
+  /// Uses the first credential found because the other credentials should have the same placement
+  /// ID for a given ad unit.
```
thanvirm: On Admob, a publisher could set up multiple mediation groups for a single ad unit and the BidMachine placement ID could be different for each mediation group. So, during signal collection, there could be multiple different BidMachine placement IDs. The actual mediation group will get resolved only after signal collection is completed and the request is sent to Admob server by the Admob SDK.

thanvirm:
Thank you for the PR. The unit tests are failing. Could you fix them? Thanks! [Here](https://github.com/googleads/googleads-mobile-ios-mediation/blob/main/AdapterUnitTests/README.md) are the instructions for running the unit tests.

